PR #412, opened 2014-04-24
Add formula for hive 0.10.0

onlynone:
* There's a backwards incompatible bug between hive 0.13.0 and earlier
   versions that make having this version around nice
 * HIVE-6050

### DIFF
--- a/hive010.rb
+++ b/hive010.rb
@@ -1,9 +1,10 @@
 class Hive010 < Formula
   homepage "https://hive.apache.org"
   url "https://archive.apache.org/dist/hive/hive-0.10.0/hive-0.10.0-bin.tar.gz"
-  sha1 "b9b9f9f1b6c5e98f2c7197ac40caaafbc4bec66a"
+  sha256 "9a99ef0545758accaa30c0ede524bcaaaaeee12b115a9ca0ebf96fa72060abee"
 
   depends_on "hadoop"
+  depends_on :java
 
   def install
     rm_f Dir["bin/*.bat"]
--- a/hive010.rb
+++ b/hive010.rb
@@ -1,4 +1,4 @@
-class Hive < Formula
+class Hive010 < Formula
   homepage "https://hive.apache.org"
   url "https://archive.apache.org/dist/hive/hive-0.10.0/hive-0.10.0-bin.tar.gz"
   sha1 "b9b9f9f1b6c5e98f2c7197ac40caaafbc4bec66a"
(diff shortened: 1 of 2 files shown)


hive010.rb
```diff
@@ -0,0 +1,26 @@
+require 'formula'
```
DomT4: :scissors: this line

hive010.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Hive < Formula
+  homepage 'http://hive.apache.org'
+  url 'http://archive.apache.org/dist/hive/hive-0.10.0/hive-0.10.0-bin.tar.gz'
```
DomT4: Double quotes everywhere, Thanks!

hive010.rb
```diff
@@ -0,0 +1,26 @@
+require 'formula'
+
+class Hive < Formula
+  homepage 'http://hive.apache.org'
```
DomT4: Both this and the ` url ` should be ` https:// ` links.

hive011.rb
```diff
@@ -0,0 +1,26 @@
+require 'formula'
```
DomT4: Same comments on this formula as the one above.

DomT4:
If this is still useful I've left some audit comments above that need addressing. If it isn't useful at this point, I'll close this at some point in the next week.

onlynone:
Just wondering why this was closed. Was there any followup to the formulas?

DomT4:
@onlynone Apologies - The lack of comment in reply and Git's way of displaying when something was first edited as opposed to when it was most recently edited threw me off here completely. I've reopened the PR and will ask the bot to take another look.

DomT4:
@BrewTestBot test this please

hive010.rb
```diff
@@ -0,0 +1,24 @@
+class Hive < Formula
```
DomT4: Both these names need changing. ` class Hive010 ` and ` class Hive011 ` respectively.

onlynone:
Should I summon @BrewTestBot , or should new commits trigger it automatically?

DomT4:
New commits trigger it automatically. It hunts for SHA changes, which a new commit causes.

DomT4:
Couple of minor comments from the test-bot and this should be okay:

```
 * Use `depends_on :java` to set JAVA_HOME
```

DomT4:
Ping.

DomT4:
Merged! Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #424, opened 2014-05-13
Update varnish3.rb

paje:
-F option to have varnishd in foreground that the LaunchRocket start/stop manager works properly


DomT4:
Is this still useful? Are there any downsides to accepting this/Will it cause problems for anything else?

paje:
Hi Tom,

for my usage it's mandatory.. 
On the dev side, Launchrocket is useful to stop/start services easily.

On a productive side, it could be problematic, but i don't know if some are using varnish a mac as a server service.. 

Best regards

PR #485, opened 2014-07-23
Premake 4.3 and 4.4 devel support.

lberezy:


DomT4:
This is available in Homebrew/Homebrew, with the same versions.

PR #491, opened 2014-07-29
Create formula for thrift v0.9.0

rafaelverger:


DomT4:
Can I get an update here? Is this still necessary/Is there still anything left to gain by accepting it/etc? Thanks!

rafaelverger:
Hey @DomT4 , I think it still not possible to install thrift v0.9.0 on mavericks without this version. The fixes/commits in homebrew repo are all related with version 0.9.1+

DomT4:
I'm a little reluctant to carry this if it's likely to break again in the future, but if you want us to consider it please fix up any comments ` brew audit --strict ` has locally, remove the head block and add an ` openssl ` dep. Thanks!

DomT4:
Ping.

rafaelverger:
Sorry for the delay @DomT4 ; I will update the code today !

DomT4:
Ping?

rafaelverger:
@DomT4 pong!

DomT4:
The patch Gists seem to have wandered off?

```
==> Downloading https://gist.githubusercontent.com/rafaelverger/58b6eeafaae7d28b06cc/raw/[ca4565122f0a1365f2409bce85dc0b8942459b18
==> Retrying download
==> Downloading https://gist.githubusercontent.com/rafaelverger/58b6eeafaae7d28b06cc/raw/[ca4565122f0a1365f2409bce85dc0b8942459b18
Error: Failed to download resource "thrift090--patch"
```

thrift090.rb
```diff
@@ -0,0 +1,78 @@
...
+      %w{[7bf1cd9deb7b483845458e901c37ad4d8404a8e7 c4419ce40b7fda9ffd58a5dad7856b64ee84e3c1b820f3a64fed0b01b4bc9c42]},
+    ].each do |hash, sha|
+      patch do
+        url "https://gist.githubusercontent.com/rafaelverger/58b6eeafaae7d28b06cc/raw/#{hash}"
+        sha256 sha
+      end
+    end
+  end
+
+  head do
```
DomT4: I'm a little meh about having a ` head ` block in this formula, given the level of patching involved, the ` HEAD ` block is likely to break much sooner rather than later.

thrift090.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  depends_on "boost"
+  depends_on :python => :optional
+
+  def install
+    system "./bootstrap.sh" unless build.stable?
+
+    exclusions = ["--without-ruby", "--without-tests", "--without-php_extension"]
+
+    exclusions << "--without-python" if build.without? "python"
```
DomT4: ` args ` is the typical wording here, rather than ` exclusions ` :).

thrift090.rb
```diff
@@ -0,0 +1,78 @@
...
+    exclusions << "--without-haskell" if build.without? "haskell"
+    exclusions << "--without-java" if build.without? "java"
+    exclusions << "--without-perl" if build.without? "perl"
+    exclusions << "--without-php" if build.without? "php"
+    exclusions << "--without-erlang" if build.without? "erlang"
+
+    ENV.cxx11 if MacOS.version >= :mavericks && ENV.compiler == :clang
+
+    # Don"t install extensions to /usr:
```
DomT4: You don't want to change the ` ' ` in ` Don't ` to a ` " `, I suspect :wink:

DomT4:
Few nits, but looking much better :+1:.

thrift090.rb
rafaelverger: Side effect of a replace all `'` by `"` :confounded:

DomT4: Heh, yeah, I know that problem well. Happens to me all the time :smile_cat:

rafaelverger:
@DomT4 check this out :)

thrift090.rb
```diff
@@ -0,0 +1,71 @@
+# The 0.9.0 tag in homebrew repository isn"t working in OSX Mavericks
```
DomT4: Replace all has caught you here. ` isn"t ` :)

thrift090.rb
```diff
@@ -0,0 +1,71 @@
...
+    # patch-remove-tr1-dependency.patch
+    %w[7bf1cd9deb7b483845458e901c37ad4d8404a8e7 c4419ce40b7fda9ffd58a5dad7856b64ee84e3c1b820f3a64fed0b01b4bc9c42],
+  ].each do |hash, sha|
+    patch do
+      url "https://gist.githubusercontent.com/rafaelverger/58b6eeafaae7d28b06cc/raw/#{hash}"
+      sha256 sha
+    end
+  end
+
+  depends_on :autoconf
```
DomT4: Should be ` depends_on "autoconf" => :build ` and the same style for automake and libtool as well; the ` :autoconf ` dep is deprecated.

thrift090.rb
```diff
@@ -0,0 +1,71 @@
...
+    system "./configure", "--disable-debug",
+                          "--prefix=#{prefix}",
+                          "--libdir=#{lib}",
+                          *args
+    ENV.j1
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/thrift --version | grep 0.9.0"
```
DomT4: Probably best not to grep for the prefix, just in case this gets updated and someone doesn't follow it up. ` system "#{bin}/thrift --version ` is fine :)

DomT4:
Nearly there :+1:

thrift090.rb
rafaelverger: `thrift --version` returns a code 1 after execution, which causes test to fail :S

DomT4: That's irritating of it. Try something like ` system "#{bin}/thrift", "--version", "2>&1" `. Might work.

DomT4:
Alright, Merged. I fixed the test locally, since it was playing up. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. In future, please note the commit message for new formulae should be styled ` thrift090 0.9.0 (new formula) ` and ideally commits squashed together, Thanks!

rafaelverger:
Awesome, @DomT4 , thanks for the tips!

PR #495, opened 2014-08-05
Add formula for ansible 1.5.4

vlad-dragos-qualitance:


DomT4:
Hi, is this still useful to accept? If so, please fix this up based on the newish guidelines on Python handling in Homebrew/Homebrew and address any comments ` brew audit --strict ` has, Thanks!

PR #513, opened 2014-09-03
Add python32: Python 3.2.3

clacke:
Python 3.2 is still supported until 2016-02, according to [PEP 392](http://legacy.python.org/dev/peps/pep-0392/).

This package is just a fork of the mainline python3.rb, with downgraded Python, installation of setuptools and pip, and two quirks at installation.

No effort has been done to allow it to coexist with python3, you have to unlink one to link the other.

### DIFF
--- a/python32.rb
+++ b/python32.rb
@@ -1,0 +1,267 @@
+require 'formula'
+
+# Python3 is the new language standard, not just a new revision.
+# It's somewhat incompatible to Python 2.x, therefore, the executable
+# "python" will always point to the 2.x version which you can get by
+# `brew install python`.
+
+class Python32 < Formula
+  homepage 'http://www.python.org/'
+  url 'http://python.org/ftp/python/3.2.3/Python-3.2.3.tar.bz2'
+  sha1 '4c2d562a0681ba27bc920500050e2f08de224311'
+  VER='3.2'  # The <major>.<minor> is used so often.
+
+  option :universal
+  option 'quicktest', 'Run `make quicktest` after the build'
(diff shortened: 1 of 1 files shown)


DomT4:
Aside the audit failures which will need addressing prior to any merge, do you have any concerns on this @tdsmith?

tdsmith:
I'm -0 on having another Python formula to maintain forever. I'd be totally thrilled to have someone set up a homebrew-[deadsnakes](https://launchpad.net/~fkrull/+archive/ubuntu/deadsnakes) tap. `pyenv` is also great.

To bring this into homebrew-versions it would have to coexist peacefully with the python3 formula and it would have to be updated against a recent python3.rb, I think.

DomT4:
@tdsmith Thanks for the feedback Tim. That's incredibly helpful for me :+1:.

Per Tim's feedback, the minimum for getting this into Versions would be for it to:

* Not ram into Python3 from Homebrew/Homebrew, presumably via Tim's suggested ` share/python32 ` solution.
* Rebase against a recent Python3 formula from Homebrew/Homebrew to get all of Tim's upstream updates there.

Otherwise this'd need to go in a special tap and couldn't be accepted here. Let us know if you want to drop the formula or want to adapt it per Tim's suggestions, Thanks!

DomT4:
Closing per inactivity. Please feel free to request reopening or open a new PR if you're still interested - If you do, please fix any comments ` brew audit --strict ` has and follow up on Tim's feedback, Thanks!

PR #526, opened 2014-09-22
Create maven321.rb

meandmax:
to install Maven 3.2.1 with brew.


DomT4:
Yes, I think Jack was suggesting it should be ` maven32 ` so the formula is pinned to the upstream branch rather than a specific ` .x ` release unless that specific ` .x ` release has some special quality needed/desired.

DomT4:
Closing per silence on this. Feel free to request reopening or open a new PR if you're still interested in getting this in. If you do either, please fix up any comments ` brew audit --strict ` has prior to submission, Thanks!

PR #549, opened 2014-10-19
Port 10.10 fixes to gcc46 and gcc47.

alexpennace:
The patch for gcc48 applies to gcc46 and gcc47, allowing those to be built on Mac OS X 10.10. Fixes #515, fixes #548.

(See commit 843d2baf2c25e3bebd6b909c7a023a14137b6d72).

Although these are old gcc versions and RVM should use a more recent gcc, it would be nice to keep these older versions available on Yosemite.

### DIFF
--- a/gcc46.rb
+++ b/gcc46.rb
@@ -44,6 +44,12 @@ class Gcc46 < Formula
   # GCC bootstraps itself, so it is OK to have an incompatible C++ stdlib
   cxxstdlib_check :skip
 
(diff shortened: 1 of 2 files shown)


sjackman:
Thanks, Jack.

JohnDDuncanIII:
for OS X 10.11, the file gcc/config/darwin-driver.c needs to be updated to include El-Capitan:
-  if (major_vers > 4 + 10)
+  if (major_vers > 4 + 11)

if not it will cause the program to segfault during that step. By including this patch (I am building this in pkgsrc on OS X 10.11), the gcc47 compilation continued uninterrupted.

PR #578, opened 2014-12-02
Elasticsearch 1.2.4 and 1.3.9

stringbean:
These are in `elasticsearch12` and `elasticsearch13` respectively.

### DIFF
--- a/elasticsearch12.rb
+++ b/elasticsearch12.rb
@@ -1,0 +1,107 @@
+class Elasticsearch12 < Formula
+  homepage "http://www.elasticsearch.org"
+  url "https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-1.2.4.tar.gz"
+  sha1 "c8bbe1f1975ffb6774744fadd0abb78616e96904"
+
+  depends_on :java => "1.7+"
+
+  def cluster_name
+    "elasticsearch_#{ENV["USER"]}"
+  end
+
+  def install
+    # Remove Windows files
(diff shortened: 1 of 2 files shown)


DomT4:
Hi! If this is still useful please address any comments ` brew audit --strict ` has locally, Thanks!

elasticsearch12.rb
```diff
@@ -0,0 +1,122 @@
+require "formula"
+
+class Elasticsearch12 < Formula
+  homepage "http://www.elasticsearch.org"
+  url "https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-1.2.4.tar.gz"
+  sha1 "c8bbe1f1975ffb6774744fadd0abb78616e96904"
+
+  depends_on :java => "1.7"
```
DomT4: ` depends_on :java => "1.7+" ` (Same for the formula below as well)

elasticsearch12.rb
```diff
@@ -0,0 +1,122 @@
...
+
+  head do
```
DomT4: :scissors: the head blocks. Versions formulae can't have ` head ` blocks.

stringbean:
@DomT4 I've made the changes you suggested and fixed the audit complaints.

DomT4:
Merged! Thanks for your contribution to Homebrew, We appreciate it! For future reference new formulae should be split into multiple commits when they don't depend on each other, and the commit reference should be ` formulaname version (new formula) `, Thanks!

PR #582, opened 2014-12-04
Fix unison232 segfault on 10.10

hightman:


DomT4:
@BrewTestBot test this please

PR #615, opened 2015-01-08
add formula for stable samba 4.1.14

bySabi:


mistydemeo:
If this builds and works consistently now, the `samba` formula can be upgraded; the main formula was only kept to samba3 because samba4 wouldn't build properly on OS X for a long time.

jacknagel:
https://github.com/Homebrew/homebrew/pull/35664#issuecomment-69528197

PR #616, opened 2015-01-08
Bumping openssl098 version from 0.9.8zc to 0.9.8zd

ace:
Openssl version was just updated upstream and was resulting in a 404 when trying to install 0.9.8zc


DomT4:
Should probably add a mirror to make sure we don't 404 in future on OpenSSL updates, given their new policy of yanking old versions as soon as new ones are available.

PR #617, opened 2015-01-10
Cmake 3.0.2

davydden:
Cmake 3.1.0 has a bug which prevents open-source software like deal.ii
from proper configuration, whereas the previous version (3.0.2)
works fine.

The added formula is CC of b950d649855977b89812ec5c3794501f073a7e09 in the main homebrew repo with bottles removed and `conflicts_with` added


davydden:
Never-mind. The problem has been fixed. Unless someone needs `3.0.2`, this can be closed.

PR #534, opened 2014-10-03
adding docker v1.1.2 version

jstenhouse:
Adding docker v1.1.2 version. This matches the older docker formula at: https://github.com/Homebrew/homebrew/blob/e48d2b50af82806b6bebece0a3657a9100a841b7/Library/Formula/docker.rb

I've tested this on MacOS Mavericks.

PR #522, opened 2014-09-16
Formula for installing Docker 1.0.1, latest with API version 1.12

markjreed:
Might be needed for non-upgraded locally-installed of docker server farms

PR #618, opened 2015-01-11
Add Docker 1.3.3 (latest compatible w/ latest CoreOS)

radeksimko:
Currently latest CoreOS stable & beta versions are only compatible with docker `1.3.3`, which is the reason for submitting this formula.

See:
screen shot 2015-01-11 at 10 16 34

https://coreos.com/releases/

I have seen both https://github.com/Homebrew/homebrew-versions/pull/534 (`1.1.2`) and https://github.com/Homebrew/homebrew-versions/pull/522 (`1.0.1`). I can't say whether it would solve the situation with CoreOS, but there should be the latest possible version compatible in the repo I'd say.

You might argue that docker is part of CoreOS, but if people want to connect to the docker daemon running in CoreOS from elsewhere (e.g. host machine to VM), they need the same version of docker.

Formula matches with https://github.com/Homebrew/homebrew/blob/21e59c6/Library/Formula/docker.rb


radeksimko:
As stated by Jenkins:
```
`require "formula"` is now unnecessary
```
so I fixed it to make build bot happy & green. ;)

DomT4:
@BrewTestBot test this please

DomT4:
Merged! Thanks for your contribution to Homebrew, We appreciate it! :smile_cat:. For future reference the commit message for new formulae should be ` docker 1.3.3 (new formula) `, Thanks!

PR #621, opened 2015-01-14
Versioned formula for libpqxx v3 (3.1.1)

activeshadow:
This formula installs libpqxx-3.1.1, and is based off the existing
formula for libpqxx.

There's a major change between version 3 and 4 that involves pqxx::prepare::treat_string. At least for some code I'm compiling on OS X, this was a show stopper. Figured having a version 3 available in Homebrew might help others out too.

### DIFF
--- a/libpqxx3.rb
+++ b/libpqxx3.rb
@@ -1,0 +1,66 @@
+require 'formula'
+
+class Libpqxx3 < Formula
+  homepage 'http://pqxx.org/development/libpqxx/'
+  url 'http://pqxx.org/download/software/libpqxx/libpqxx-3.1.1.tar.gz'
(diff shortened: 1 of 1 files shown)


DomT4:
@BrewTestBot test this please

DomT4:
Could you possibly add a ` test do ` to the formula? See cmake.rb for an example of an application formula with a good test and tinyxml2.rb for an example of a library formula with a good test. Thanks!

DomT4:
@BrewTestBot test this please

DomT4:
Merged! Thanks for your contribution to Homebrew, we appreciate it! In future please note the preferred commit message for new formulae is ` example1 1.2.3 (new formula) ` and a ` test do ` block would be awesome, even something simple. Thanks!

PR #623, opened 2015-01-15
llvm35 3.5.1

lembacon:


llvm35.rb
```diff
@@ -106,10 +110,13 @@ class Llvm35 < Formula
...
-  depends_on 'gmp'
-  depends_on 'libffi' => :recommended
+  depends_on "gmp"
+  depends_on "libffi" => :recommended
 
-  def ver; '3.5'; end # version suffix
+  # version suffix
+  def ver
+    version.to_s.slice(/\d+\.\d+/)
+  end
```
jacknagel: This can continue to be a static string, since this formula will never be upgraded beyond 3.5.

lembacon: Fixed in 18b3d22.

PR #629, opened 2015-01-25
openssl101 1.0.1l

DomT4:
This is being EOL’ed at the end of this year alongside the 0.9.8 branch, but it may be useful in the mean time.


DomT4:
@BrewTestBot test this please

PR #636, opened 2015-01-29
Add Play 1.3.0

ghostganz:
Adding the "new" stable Play 1.3.0 as "play13"

### DIFF
--- a/play13.rb
+++ b/play13.rb
@@ -1,0 +1,14 @@
+class Play13 < Formula
+  homepage "http://www.playframework.org/"
+  url "http://downloads.typesafe.com/play/1.3.0/play-1.3.0.zip"
+  sha1 "0637fc40bc98675f56d3f408e73be6e6c067c789"
+
+  def install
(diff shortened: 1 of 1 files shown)


play13.rb
```diff
@@ -0,0 +1,16 @@
+require 'formula'
```
DomT4: This line can be snipped.

play13.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class Play13 < Formula
+  homepage 'http://www.playframework.org/'
```
DomT4: Double quotes everywhere please, Thanks!

ghostganz:
@DomT4 Thanks, I've fixed the quotes and the require now.

DomT4:
Thanks! Any chance of a small test? See cmake.rb for an example of an application formula with a good test and tinyxml2.rb for an example of a library formula with a good test. Thanks!

ghostganz:
I only ported it from the play 1.2.x recipe, I'll leave it for someone with more spare time to create a test. :)

PR #643, opened 2015-02-11
Add swig 3.0.4

destijl:
This is the same as the 3.0.5 file in the main branch, minus the bottles and
with a different link and hash.

Why: M2Crypto fails to build with swig 3.0.5, which breaks a number of projects.

### DIFF
--- a/swig304.rb
+++ b/swig304.rb
@@ -1,0 +1,41 @@
+class Swig304 < Formula
+  homepage "http://www.swig.org/"
+  url "https://downloads.sourceforge.net/project/swig/swig/swig-3.0.4/swig-3.0.4.tar.gz"
+  sha1 "088b2fb1f3167703f79fb187fdb3b80513426eb1"
+
+  option :universal
+
+  depends_on "pcre"
+
+  def install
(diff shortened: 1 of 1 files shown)


destijl:
Hi @jacknagel, just a friendly ping :)  FTR I've filed an issue with M2Crypto to get the actual problem fixed:
https://github.com/M2Crypto/M2Crypto/issues/24

lucianocheng:
Just hit this issue myself (https://github.com/georgebrock/1pass/issues/21).  For anyone who wants to use this workaround, you can install this from @destijl 's PR directly by running

```
brew install https://raw.githubusercontent.com/destijl/homebrew-versions/master/swig304.rb
```

Also thanks @destijl for coming up with a workaround for this problem, every one of our OSX dev environments croaked when we `brew updated` to swig 3.0.5

DomT4:
@BrewTestBot test this please

DomT4:
Merged! Thanks for your contribution to Homebrew, It is appreciated! :smile_cat:. For future reference new formula's commit reference should be ` swig 3.0.4 (new formula) ` and similar, Thanks!

PR #647, opened 2015-02-12
mass culling

DomT4:
Proposal for culling an extensive list of formulae with the general desire of making Homebrew/Homebrew-versions less… unwieldy. Pretty much expect this to be controversial and probably ultimately, rejected - But I’m hoping it at least gets the ball rolling a little on putting a stricter policy in place on what Homebrew/Versions accepts and why.

The reasons for culling these particular formulae weren’t random, but I didn’t really think anyone would appreciate a lengthy list detailing the removals. Essentially, you can pigeonhole the removals into one or several of three categories; Upstream severe deprecation, Security issues or Formulae that nobody has touched in literally years, even to resolve audit issues or add basic functionality tests - Which are now basically mandatory in Homebrew/Homebrew.

Happy to discuss individual formulae as desired.


mikemcquaid:
I don't think we want to do this. People want to stick on old versions indefinitely. The problem with this tap is that insufficient numbers of @Homebrew/owners care about this tap enough to watch the PRs.

mikemcquaid:
> The problem with this tap is that insufficient numbers of @Homebrew/owners care about this tap enough to watch the PRs.

(and I'm definitely one of them)

DomT4:
> People want to stick on old versions indefinitely.

My (entirely personal) viewpoint on this is that it wouldn't be unreasonable for Homebrew to 'inflict' basic formulae acceptance/support rules in the style of Acceptable Formulae.md in Homebrew/Homebrew. Homebrew already asks people to maintain their own taps for niche formulae, and a lot of this stuff has zoomed past niche and into archaic. That's at least where I'm coming from on this sort of thing, as well as a general desire to make Homebrew/Homebrew owners less tired of having to deal with outdated versions and supporting them.

I'll close this though as it sets some discussion up, and people will just freak if they see it.

PR #650, opened 2015-02-12
Adding phantomjs 1.9.8 as an option

hootgabriel:
First formula, let me know if anything else needs to be done (I ran the audit command and it checks out)

### DIFF
--- a/phantomjs198.rb
+++ b/phantomjs198.rb
@@ -1,0 +1,16 @@
+class Phantomjs198 < Formula
+  homepage 'http://www.phantomjs.org/'
+  url 'https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-macosx.zip'
+  sha1 'd70bbefd857f21104c5961b9dd081781cb4d999a'
+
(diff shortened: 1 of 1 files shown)


brrygrdn:
:+1: for this, let me know if there's anything I can do to help

phantomjs198.rb
```diff
@@ -0,0 +1,16 @@
+class Phantomjs198 < Formula
+  homepage 'http://www.phantomjs.org/'
+  url 'https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-macosx.zip'
```
DomT4: We don't use upstream binaries; We produce our own from the source (bottles). We'll need to build it from source.

phantomjs198.rb
```diff
@@ -0,0 +1,16 @@
+class Phantomjs198 < Formula
+  homepage 'http://www.phantomjs.org/'
```
DomT4: Double quotes everywhere please, Thanks!

PR #651, opened 2015-02-14
node 0.10.36

xaka:
It's based on latest formula from https://github.com/Homebrew/homebrew/blob/06b10bc64605ece33b56d2607a1480cf8177fd8f/Library/Formula/node.rb right before it had been bumped to 0.12.x

### DIFF
--- a/node010.rb
+++ b/node010.rb
@@ -1,0 +1,130 @@
+# Note that x.even are stable releases, x.odd are devel releases
+class Node010 < Formula
+  homepage "https://nodejs.org/"
+  url "https://nodejs.org/dist/v0.10.36/node-v0.10.36.tar.gz"
+  sha256 "b9d7d1d0294bce46686b13a05da6fc5b1e7743b597544aa888e8e64a9f178c81"
+
+  deprecated_option "enable-debug" => "with-debug"
+
+  option "with-debug", "Build with debugger hooks"
+  option "without-npm", "npm will not be installed"
+  option "without-completion", "npm bash completion will not be installed"
+
+  depends_on :python => :build
+  depends_on "openssl" => :optional
+
+  fails_with :llvm do
+    build 2326
+  end
+
+  resource "npm" do
+    url "https://registry.npmjs.org/npm/-/npm-2.5.1.tgz"
+    sha1 "23e4b0fdd1ffced7d835780e692a9e5a0125bb02"
+  end
+
+  def install
+    args = %W[--prefix=#{prefix} --without-npm]
+    args << "--debug" if build.with? "debug"
+
+    if build.with? "openssl"
+      args << "--shared-openssl"
+    else
+      args << "--without-ssl2" << "--without-ssl3"
+    end
+
+    system "./configure", *args
+    system "make", "install"
+
+    if build.with? "npm"
+      resource("npm").stage buildpath/"npm_install"
+
+      # make sure npm can find node
+      ENV.prepend_path "PATH", bin
+
+      # make sure user prefix settings in $HOME are ignored
+      ENV["HOME"] = buildpath/"home"
+
+      # set log level temporarily for npm's `make install`
+      ENV["NPM_CONFIG_LOGLEVEL"] = "verbose"
+
+      cd buildpath/"npm_install" do
+        system "./configure", "--prefix=#{libexec}/npm"
+        system "make", "install"
+      end
+
+      if build.with? "completion"
+        bash_completion.install \
+          buildpath/"npm_install/lib/utils/completion.sh" => "npm"
+      end
+    end
+  end
+
+  def post_install
+    return if build.without? "npm"
+
+    node_modules = HOMEBREW_PREFIX/"lib/node_modules"
+    node_modules.mkpath
+    npm_exec = node_modules/"npm/bin/npm-cli.js"
+    # Kill npm but preserve all other modules across node updates/upgrades.
+    rm_rf node_modules/"npm"
+
+    cp_r libexec/"npm/lib/node_modules/npm", node_modules
+    # This symlink doesn't hop into homebrew_prefix/bin automatically so
+    # remove it and make our own. This is a small consequence of our bottle
+    # npm make install workaround. All other installs **do** symlink to
+    # homebrew_prefix/bin correctly. We ln rather than cp this because doing
+    # so mimics npm's normal install.
+    ln_sf npm_exec, "#{HOMEBREW_PREFIX}/bin/npm"
+
+    # Let's do the manpage dance. It's just a jump to the left.
+    # And then a step to the right, with your hand on rm_f.
+    ["man1", "man3", "man5", "man7"].each do |man|
+      # Dirs must exist first: https://github.com/Homebrew/homebrew/issues/35969
+      mkdir_p HOMEBREW_PREFIX/"share/man/#{man}"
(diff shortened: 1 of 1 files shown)


DomT4:
Any particular reason you want Homebrew to keep supporting the 0.10.x branch? The ` devel ` and ` head ` blocks would need to be killed here as well.

xaka:
I've removed both `devel` and `head` blocks. 

The only reason I'd like to have it is I'm using it now and I cannot switch to 0.12 nor I want to switch to `nvm` for now. And as there are 0.6 and 0.8, why not to have  0.10 after all? :)

DomT4:
> I've removed both devel and head blocks.

Thanks, Appreciated.

> And as there are 0.6 and 0.8, why not to have 0.10 after all? :)

To be fair, If it was my choice I wouldn't support 0.6 or 0.8. I can see a justification for 0.10 if absolutely necessary, but I loathe supporting versions that upstream have long since abandoned by upstream and pose potential security issues. Thankfully the choice is not mine :smile_cat:

xaka:
I totally agree with you here. 0.6.x is no longer supported and when 0.12 came out, 0.8.x stopped being supported as well, so there is no reason to keep updating formulas. For 0.10.x it's different as it's not abandoned yet and we probably gonna see couple more releases of it.

Ambroos:
There are some changes in 0.12 that break certain projects, I ran into this. Having 0.10 in homebrew/versions would be a welcome addition for a while.

Romanior:
Plus one here, check out please for instance this node-sass incompatibility.
https://github.com/sass/node-sass/issues/550

DomT4:
No need for +1s :smiley_cat:, the maintainers are likely to accept this when they see it.

xaka:
@DomT4 I think voting is very important (and actually one of biggest missing features of GitHub) for many reasons:

1. It answers on your first question - "any particular reason to keep supporting the 0.10.x branch". It's because community needs it for now. Usually it's not easy to get community involved so even 2 people is a good start.
2. It helps maintainers to stay focus on what is important for community. Otherwise why would you open source your pet and allow people to send PRs? I assume because you wanted to be community-driven. It doesn't mean community is always right, but at least you know that among XXX PRs, people really need A, B and C at this moment.
3. Maintainers may not accept it nor provide any feedback besides "doesn't make sense", but now they see people are gathering around, so they are expected to give or suggest a workaround, or even change their mind.

DomT4:
I already had 1 answered by you and Roman to an extent I'm happy with. I don't maintain this particular repository (Or Homebrew/Homebrew) so the maintainers may want more, but I'm content it's more than just "Support x old version because that's what we always do".

As for 2 & 3, the maintainers in Homebrew/Homebrew tend to specifically request people *don't* add votes to an issue or PR. Although I can't read their minds that's potentially because it then makes dealing with that issue/PR a complete nightmare as you have to wade through a billion votes to find the substance you need to resolve the issue/PR one way or another.

I don't think the maintainers are likely to forget the wider community suddenly, Mike has been on a one-man mission the past 2 days to move Homebrew's bottles away from Sourceforge to a new provider which is easily the most anticipated fix/change in Homebrew/Homebrew's issues and PRs for quite some time. It's a huge amount of work but you can see 20+ commits upstream in Homebrew/Homebrew on that change alone for the past 24 hours or so.

This will come to the attention of the maintainers sooner or later, and you'll get maintainer feedback on it. That will happen regardless of voting :smiley:

xaka:
Just to be clear - I'm not saying that voting means "we need maintainers right here and right now", it just means "whenever you guys have time, please take a look here first". We all appreciate a hard work and no one is in power to command what to do. For sure.

node010.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  deprecated_option "enable-debug" => "with-debug"
+
+  option "with-debug", "Build with debugger hooks"
+  option "without-npm", "npm will not be installed"
+  option "without-completion", "npm bash completion will not be installed"
+
+  depends_on :python => :build
+
+  # Once we kill off SSLv3 in our OpenSSL consider making our OpenSSL
```
DomT4: Kill this section and replace it with ` depends_on "openssl" => :optional `.

node010.rb
```diff
@@ -0,0 +1,130 @@
...
+  # an optional dep across the whole range of Node releases.
+
+  fails_with :llvm do
+    build 2326
+  end
+
+  resource "npm" do
+    url "https://registry.npmjs.org/npm/-/npm-2.3.0.tgz"
```
DomT4: ` url "https://registry.npmjs.org/npm/-/npm-2.5.1.tgz" `

node010.rb
```diff
@@ -0,0 +1,130 @@
...
+    sha1 "3588ec5c18fb5ac41e5721b0ea8ece3a85ab8b4b"
```
DomT4: ` sha1 "23e4b0fdd1ffced7d835780e692a9e5a0125bb02" ` (Re above change).

node010.rb
```diff
@@ -0,0 +1,130 @@
...
+  end
+
+  def install
+    args = %W[--prefix=#{prefix} --without-npm]
+    args << "--debug" if build.with? "debug"
+    args << "--without-ssl2" << "--without-ssl3"
```
DomT4: Please replace the last ` args ` line here with [this block](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/node.rb#L43-L47) of code.

node010.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    if build.with? "npm"
+      resource("npm").stage buildpath/"npm_install"
+
+      # make sure npm can find node
+      ENV.prepend_path "PATH", bin
+
+      # set log level temporarily for npm's `make install`
+      ENV["NPM_CONFIG_LOGLEVEL"] = "verbose"
+
```
DomT4: Please add [these two lines](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/node.rb#L57-L58).

DomT4:
Requested a bundle of changes, just syncing with changes we made in Homebrew/Homebrew's Node in the recent update (other than the version bump obviously) that iron out certain bugs/issues/etc going forwards.

After that, I'm happy to merge this in; There may be a slight hold up whilst Homebrew/Homebrew resolves some bottling issues with new formulae, but I'll merge this in as soon as we resolve that.

node010.rb
```diff
@@ -0,0 +1,130 @@
+# Note that x.even are stable releases, x.odd are devel releases
+class Node010 < Formula
+  homepage "https://nodejs.org/"
+  url "https://nodejs.org/dist/v0.10.36/node-v0.10.36.tar.gz"
+  sha256 "b9d7d1d0294bce46686b13a05da6fc5b1e7743b597544aa888e8e64a9f178c81"
+
+  bottle do
```
DomT4: Oh, scrap this ` bottle do ` block as well. I'll be creating new bottles when I pull this.

xaka:
All requested changes have been applied.

DomT4:
Thanks!

DomT4:
Merged! Thanks for your contribution to Homebrew, We appreciate it! For future reference, preferred commit message for new formula is ` node 0.10.36 (new formula) `. Thanks! :smiley_cat:

xaka:
Got it, sir. Thank you too! Now it's time to exhale :)

PR #652, opened 2015-02-14
Update gmp4 URLs

ghg:
The current first two URLs point to ftp endpoints, which are blocked for some people (well, at least me).

This PR uses the gnu.org HTTP URL as primary. The anl.gov URL is deleted
as it is no longer available.

% curl -i http://mirror.anl.gov/pub/gnu/gmp/gmp-4.3.2.tar.bz2
HTTP/1.1 403 Forbidden

After pushing, I noticed some whitespace changes. I can put these into a separate commit if needed.

Thanks!
Greg


DomT4:
Audit failures:

```
==> brew audit gmp4							 FAILED
Error: 2 problems in 1 formulae
gmp4:
 * "ftpmirror.gnu.org" is preferred for GNU software (url is http://ftp.gnu.org/gnu/gmp/gmp-4.3.2.tar.bz2).
 * ftp.gnu.org urls should be https://, not http:// (url is http://ftp.gnu.org/gnu/gmp/gmp-4.3.2.tar.bz2).
```

If you can, remove the whitespacing from the patch as well. It doesn't seem to have broken anything in this case, but whitespacing in patches is a delicate and extremely prone to breakage thing, sadly.

PR #644, opened 2015-02-11
gnupg21 2.1.2

DomT4:
Bump.

PR #653, opened 2015-02-14
Update gnupg21 to 2.1.1

Midar:


DomT4:
Dupe of #644.

Midar:
Uses SHA256 instead of SHA1, though (SHA1 is getting worryingly weak).

DomT4:
Upstream display SHA1 on their website, which is why we keep using it. Whatever upstream display, unless it's MD5, we use - It allows people to easily independently verify integrity. If SHA1 breaks, Git is screwed, and if Git is screwed, Homebrew's security would be broken anyway. By all means, nag upstream to update their displayed checksums to something stronger, I would entirely welcome that :wink:

Midar:
If that's the reasoning, there's also ftp://ftp.gnupg.org/gcrypt/gnupg/gnupg-2.1.2.tar.bz2.sig ;)

DomT4:
Homebrew doesn't support GPG verification at the moment, although it is something I'd like to work on long-term.

Midar:
Ah, right, that was homebrew-cask.

Still, do we really want to rather use a digest algorithm that is not that far from being weak? I think it's more secure to download the tarball, verify the signature and store the SHA256 in the Formula - which is what I did.

DomT4:
Yup. Cask have implemented a ` gnupg ` check. It's a difficult thing to screw up though, the whole ` GPG ` system is, and so far Homebrew/Homebrew's maintainers have been reluctant to force a GnuPG dependency on everyone.

> Still, do we really want to rather use a digest algorithm that is not that far from being weak?

Unless you're the NSA, or a state-level equivalent, there are no practical attack of SHA1 at this time. You'd need to spend somewhere near $3m to break *one* SHA1 checksum. 

I agree that we *should* be moving to SHA256, better safe than sorry, and it's more future-proof - But most Linux distros are only just using SHA1, and most are still supporting MD5, and given GnuPG display the SHA1 checksums on their website moving away from that would actually reduce people's ability to independently verify the checksums as desired. I'd love to move away from SHA1, but I don't think the argument is a convincing enough one for Homebrew's maintainers yet *(There have been several issues about this at Homebrew/Homebrew in the last 6 months or so. The discussion hasn't gone anywhere new in that time that has seemed to convince the maintainers any further)*.

Midar:
> Unless you're the NSA, or a state-level equivalent, there are no practical attack of SHA1 at this time. You'd need to spend somewhere near $3m to break one SHA1 checksum.

Exactly! GnuPG is one of the few things where it would actually be beneficial for them to forge the SHA1 and well worth the $3m. For other packages, this is not really an issue, agreed.

DomT4:
If the NSA were going to blow $3 on forging one SHA1 checksum, It wouldn't be on Homebrew. If they were going to compromise GnuPG they'd compromise the upstream checksums - Which would get everyone, not just a tiny section of OS X users.

They'd have further issues because the vast majority of people who will install anything through Homebrew will use bottles, which have their own unique SHA1 per operating system - So you'd basically have to compromise the source SHA1, and then the 3 bottle SHA1 sums. That's $12mil to catch a tiny proportion of OS X users. Even the NSA, with their ridiculous funding, would be pretty stupid to throw away $12mil on a tiny, tiny proportion of OS X users.

Midar:
> If the NSA were going to blow $3m on forging one SHA1 checksum, It wouldn't be on Homebrew.

That's non-sense. If they create a new tarball with the same SHA-1 checksum, everybody's affected, not only homebrew. Everybody whom they can MITM and builds from source is affected.

mikemcquaid:
Dupe as said earlier.

PR #655, opened 2015-02-15
Add Play Framework 1.3

fdrouet:


play13.rb
```diff
@@ -0,0 +1,18 @@
...
+    rm_rf "python" # we don't need the bundled Python for windows
+    rm Dir["*.bat"]
+    libexec.install Dir["*"]
+    bin.mkpath
+    chmod 0755, libexec+"play"
+    ln_s libexec+"play", bin
+  end
+
+  test do
+    # no test yet
```
DomT4: Could you possibly add a test down here if one is possible? If there isn't anything that can be done, remove the empty ` test do ` block. Thanks!

fdrouet: If I don't put test block, the jenkins job which check the formula syntax raise an error.
That is why I had to add one

DomT4: We can ignore the test bot's failure, if that's the only thing that fails.

DomT4:
@BrewTestBot test this please

PR #656, opened 2015-02-16
Update glfw3.rb to GLFW 3.1.

Geenz:
I've updated the version of GLFW3 in glfw3 to GLFW 3.1.  GLFW is now using GitHub, so I figured it'd be a good idea to switch the download to pull from GitHub instead of SourceForge as well.  I've tested this locally, everything seems to work.


Geenz:
Oh, there's already a PR for this! My bad.

PR #660, opened 2015-02-17
Update appledoc22 to version 2.2.1

lumaxis:


mistydemeo:
Thanks for the contribution! However, this should continue to build from source rather than install the prebuilt binary.

lumaxis:
Is building from source always the preferred method? Because the current implementation broke - probably with some system change.
Also, the way this formula is setup, it would pull down a pre-set source code version and just build that.
I don't see any real benefit in that but instead just unnecessary complexity and fragility but am happy to hear other opinions :smiley:

lumaxis:
Updated the PR, should be mergable now. Thanks again.

mistydemeo:
Building from source is always the preferred method, yes. Homebrew's a source-based package manager that also distributes binaries that we build, using the source recipes; all of the packages in this repository need to be built from source.

>  Also, the way this formula is setup, it would pull down a pre-set source code version and just build that.

This is the standard for Homebrew! Formulae are set up to use specific versions.

mistydemeo:
Thanks for updating the PR so promply!

mistydemeo:
@BrewTestBot test this please

DomT4:
@mistydemeo Any objections to this being merged? Had the bot regenerate the bottles this morning so they're available on Bintray rather than SF, so one of us can pull it if you're happy :)

DomT4:
Merged! Thanks for your contribution to Homebrew, We appreciate it!

PR #662, opened 2015-02-20
Apply 10.10 patches to gcc43 and gcc44

lethosor:


jacknagel:
Sorry, but the lack of 10.10 support in these formulae is intentional.

lethosor:
If you mean that these formulae don't compile under 10.10, that's what these patches are intended to address (similar to ec847130 for gcc45). I'm not sure about gcc43, but a user on IRC reported that gcc44 compiles and works under 10.10 with this patch.

jacknagel:
I mean we are intentionally not supporting gcc43 and gcc44 on 10.10+ going forward.

dpo:
@jacknagel Why is that? That would be too bad because in the scientific world, many use Matlab, and the MathWorks (who commercialize Matlab) only officially support gcc/gfortran 4.3 on OSX: http://www.mathworks.com/support/compilers/R2015a/index.html?sec=maci64. Although they plan to drop support for gfortran altogether in the future, many users still have older releases of Matlab.

jacknagel:
Given the resources we have available and the age of these versions (the last release of GCC 4.4 was three years ago, and the last release of 4.3 was nearly four years ago), it doesn't make sense for us to continue to maintain them.  We are still maintaining five versions: 4.5, 4.6, 4.7, 4.8, and 4.9 (which is still too many in my opinion, again, given the resources we have available).

If there is a need, I would suggest interested parties maintain them outside of this repository.

PR #665, opened 2015-02-23
node04: build adjustments

DomT4:
This formula is completely dead on my machine. Nothing Homebrew related, It just won't build any more. Fails on one of the internal deps, with a *long* list of compile warnings.

If it fails on all three build bots, We can either MMOS Req it or boneyard it completely, I guess.


DomT4:
Yup, fails. Any thoughts on boneyard/deletion/MMOS Req since we can't do anything with this on any of the three systems we can test @jacknagel @mikemcquaid?

jacknagel:
>The build-tool has reset ENV. --env=std required.

Try `env :std`? Failing that, boneyard it.

DomT4:
Gave it a whirl locally without any luck. Have pushed here to see if it fixes the failure on <10.10 at all. Thanks for the feedback! :)

DomT4:
Colour me surprised, that fixed the 10.8 build. Guess we can MMOS Req it then.

PR #672, opened 2015-02-24
node08: openssl fixes and modernisation

DomT4:
If we have to maintain old Node versions, at least this way they aren't a nightmare in difference from Homebrew/homebrew and they are secure compared to the bundled OpenSSL versions, or in some cases, linking into the system OpenSSL.


DomT4:
Any objections on this @jacknagel @mikemcquaid? It's recent enough to be tugged into line with the Homebrew/homebrew formula, but it's a pretty drastic rewrite.

It'll mean less trouble going forwards, as the 08 formula was missing a whole bunch of important changes we'd pushed to Homebrew/homebrew in the last 6+ months. It also enforces OpenSSL for obvious reasons here. Just thought given it was a more major change than just adding OpenSSL dep and revisioning I'd run it past you both.

There's nothing too radical as such. It's essentially the Homebrew/Homebrew formula, just with an old Node version and an enforced OpenSSL dep.

mikemcquaid:
:+1: with me if the build is green and you think it will save us effort longer-term.

DomT4:
Yeah, just means we can keep it in relative sync with the Homebrew/Homebrew formula going forwards rather than having it so specialised and difficult to maintain. The Node formula in Homebrew/homebrew is actually relatively stable thankfully, so backporting those changes should save some work going forwards. Have merged it in.

PR #677, opened 2015-02-25
Create phantomjs198

BPScott:
PhantomJS 1.9.8 is the last release prior to PhantomJS 2.0, which is missing some backwards compatible functionality so some projects still rely on the 1.9.x version.

Formula taken from https://github.com/Homebrew/homebrew/blob/9089836d7539aa5530cbfc6cc3d9e3772ecc78b3/Library/Formula/phantomjs.rb


phantomjs198.rb
```diff
@@ -0,0 +1,32 @@
+require "formula"
```
DomT4: :scissors: this line.

phantomjs198.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class Phantomjs198 < Formula
+  homepage "http://www.phantomjs.org/"
+
+  stable do
```
DomT4: And kill this block and just make it the usual:

```
homepage
url
sha256
```

phantomjs198.rb
```diff
@@ -0,0 +1,32 @@
...
+    url "https://github.com/ariya/phantomjs/archive/1.9.8.tar.gz"
+    sha256 "3a321561677f678ca00137c47689e3379c7fe6b83f7597d2d5de187dd243f7be"
+  end
+
+  depends_on "openssl"
+
+  def install
+    if build.stable? && MacOS.prefer_64_bit?
```
DomT4: We're always building stable here, so you can just change this to ` if MacOS.prefer_64_bit? `

DomT4:
Few nits. Nothing huge. Happy to merge when nits are addressed.

BPScott:
@DomT4 thanks for the feedback. I've addressed the nits and pushed a new commit.

Looks like the build failure is due to a `brew doctor` [regression in the previous build](http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/249/version=mavericks/testReport/brew-test-bot/mavericks/) that has hung around for reasons I don't understand.

BPScott:
I've just spotted that this is a dupe of #650

DomT4:
> I've addressed the nits and pushed a new commit

Thanks, Appreciated!

>  has hung around for reasons I don't understand.

It's fine, I can ignore this. Everything else worked as expected so we're cool to ignore the bot failure here.

> I've just spotted that this is a dupe of #650

The other one isn't going to get merged in its current form; I'm happy to accept yours' instead :+1:.

BPScott:
Cool cool. Anything more for me to do here?

DomT4:
Nope, merged :). Ideally commit messages for new formulae in future should be styled ` phantomjs198 1.9.8 (new formula) ` but I fixed that here, so no worries. Thanks for your contribution to Homebrew!

PR #679, opened 2015-02-26
appledoc20: modernise clang failure

DomT4:


DomT4:
This is super-likely to fail the build bot, but I wanted to modernise the clang failure regardless.

DomT4:
@BrewTestBot test this please

DomT4:
These two appledoc PRs aren't going to pass the bot for hell or high water without an MMOS Requirement, but I don't really want to give them one without actually having anything older than 10.8 to test on, so I'll just merge the style changes for now and we can review the situation in 6 months or so. They already don't build on OS X 10.8, 10.9, 10.10 confirmed, so I'm not breaking anything with the style fixes.

PR #682, opened 2015-02-26
gcc43: modernise

DomT4:
Modernises the gcc43 formula. My underlying motivation is to generate bottles for all the gcc formulae that we can, to save people the four and a half year compile time of gcc from source.


DomT4:
@BrewTestBot test this please

PR #690, opened 2015-02-28
gcc43: basic style tweaks

DomT4:
Basic style fixes with the underlying motivation being bottling this.


DomT4:
Always forget how delicate ` gcc `'s compile is, heh. Nearly there. The latest commit should fix the latest and last failure.

PR #698, opened 2015-03-01
Fix mountain lion tcl issues for postgresql formulas

sigerello:
@DomT4 Here is fixes for all postgresql versions except 8. I'll add it later after @BrewTestBot finish it.


sigerello:
Or should I make separate PRs for each version?

DomT4:
Individual commits per formula is preferred, but in the same PR is fine. Thanks for doing this! :+1:

DomT4:
Merged, Thanks!

PR #702, opened 2015-03-07
New formula: solr4

arnested:
Now that the `solr` formula in the master homebrew repos is up to version 5 it will be useful to have a `solr4` version around.


solr4.rb
```diff
@@ -0,0 +1,43 @@
+require "formula"
```
DomT4: :scissors: this line and this should be okay :+1:

arnested: Removed the require, added Java dependency, and added a naïve test.

DomT4:
Merged! Thanks for your contribution to Homebrew, we appreciate it :smiley_cat:. In future tests should be a little more thorough, but I let it go this time because Java stuff is a complete nightmare to test.

PR #638, opened 2015-02-02
erlang-r14: Incorporate patch from Erlbrew project

L2G:
This fixes build errors with llvm-gcc and clang.

PR #704, opened 2015-03-07
Compile erlang r16 with openssl from homebrew.

lindmartin:
Added "--with-ssl=#{Formula["openssl"].opt_prefix", this is identical to the argument in the  erlang formula  (https://github.com/Homebrew/homebrew/blob/master/Library/Formula/erlang.rb#L67) in the homebrew repository.

Tested on Yosemite.


DomT4:
Does it not currently use Homebrew's ` openssl `? I notice the dep is there, so it should be, even without the explicit specification.

lindmartin:
If I remove that line "--with-ssl=..." and issue "brew reinstall erlang-r16".
When running "crypto:start()." in erlangs shell I get an error "** exception error: undefined function crypto:start/0".
When looking in /usr/local/Cellar/erlang-r16/erlang-r16/R16B03-1_1/lib/erlang/lib/ i don't see any file named crypto.

When running the brew reinstall with --verbose the configure script says that it will compile without ssl.
*********************************************************************
**********************  APPLICATIONS DISABLED  **********************
*********************************************************************

crypto         : No usable OpenSSL found
ssh            : No usable OpenSSL found
ssl            : No usable OpenSSL found

DomT4:
Thanks, I'll look into this later tonight. Looks good to me though.

There's an ongoing issue with Erlang testing around Erlang storing its compile path internally, which makes it difficult for us to bottle unfortunately, so when I run through this and #638 later tonight I may well have some requests around adding ` inreplace `s in various places post-compile to allow us to bottle things. If we can't fix the bottle situation though I'll just merge this in and we'll not support bottles on the Erlang formulae for a while longer.

DomT4:
I've merged this in bottleless for now. Ironing out the bottle situation is taking a little while, testing so far has been a tad fruitless, and the SSL fix is important enough not to wait. I'll generate bottles for the formula as soon as I can. Thanks for your contribution to Homebrew, We appreciate it! :smiley_cat:

PR #705, opened 2015-03-07
solr14 and solr36: cleanup formulas

arnested:
Based on my findings from #702 I cleaned up the `solr14` and `solr36` formulas:

- Prefer double-quoted strings
- `require "formula"` is now unnecessary
- Stable: SHA1 checksums are deprecated, please use SHA256
- A `test do` test block should be added (although a very trivial one)
- Added `depends_on :java`
- archive.apache.org urls should be https://, not http


DomT4:
Merged! Thanks again! :smiley_cat:

PR #710, opened 2015-03-10
Compile Erlang with openssl

goofansu:
Module "crypto" is absent without ssl.

Applications such as rebar need crypto module.


DomT4:
Dupe of #704, which will be merged soon.

PR #712, opened 2015-03-12
mysql51: modernise, move to homebrew openssl by default, audit fixes

DomT4:
Modernises the formula, adds a test, moves to Homebrew's OpenSSL by default (Although still providing an opt-out, for now), passing the `brew audit --strict` test. Will bottle as well.


dnrce:
@DomT4, is there a reason `--with-partition` was removed from the build config as part of this change?

DomT4:
Rats. Human error. Sorry, will fix that shortly.

DomT4:
Human error = I screwed up :wink:

DomT4:
Do a ` brew install mysql51 ` after #751 is merged and you'll have it back. Apologies again.

dnrce:
Awesome.  Thanks for the very speedy fix!

DomT4:
Should be good now :+1:

PR #713, opened 2015-03-13
subversion16: modernise

DomT4:
The deps url is dead, and the universal build options are no longer recognised by Homebrew.

Tried to replace the former with independent vendoring of each resource as per dictated by the get-deps script in this subversion version, and including Homebrew deps for the other elements.

Replaced the latter with modernised language.


DomT4:
This might be a lost cause given it's difficult to test the new layout on the old systems, but semi-amazingly it will actually build even on 10.10.3 with the changes. Also, that we haven't had *any* reports that either the deps link is dead or the ` build_universal? `/`build.universal?` language is wrong, so perhaps there aren't that many users and it'd be better to boneyard.

Thoughts welcome from @Homebrew/owners, particularly Misty, given your experience with the older OS X versions.

mikemcquaid:
:+1: from me if it works on older versions still.

DomT4:
Without further maintainer feedback I'll probably merge this in 5 days or so. I don't have an old-old VM to test on, I tend to defer to Misty for the older OS X versions, but given the deps url has gone the way of the dodo merging this would be less broken than the current formulae even if there are kinks to iron out later. Right now, the current formula errors out on the download 100% of the time... It's hard to break it more than that.

DomT4:
I'll drop the revision if we can't get a test on an older machine though.

mikemcquaid:
@DomT4 sounds good to me :+1:

DomT4:
Revision killed per lack of verification. Have kicked the bot off on a new build, after which I'll merge. Shouldn't be any problems with the formula, we're essentially doing what we did before, just in a less direct/easy way, but I'll tweak things if they're reported to be broken on the older OS X versions post-merge.

DomT4:
We have :rocket:. Merged in https://github.com/Homebrew/homebrew-versions/commit/22abc962457f40bba27f01beb40b6f4345dc15a1

PR #714, opened 2015-03-14
add pandoc 1.13.1

randy3k:
this is required when installing http://www.rstudio.com/products/rstudio/download-server/
also related to https://github.com/Homebrew/homebrew/pull/37668


pandoc1131.rb
```diff
@@ -0,0 +1,28 @@
...
+  def install
+    cabal_sandbox do
+      cabal_install "--only-dependencies"
+      cabal_install "--prefix=#{prefix}"
+    end
+    cabal_clean_lib
+  end
+
+  test do
+    system "pandoc", "-o", "output.html", prefix/"README"
```
DomT4: Should this be ` system bin/"pandoc" ` or? I notice the upstream Homebrew/homebrew formula has the layout this way, without the ` bin ` so it may well be correct.

randy3k: Since it passes the test, I think it is okay to just do `pandoc`.

DomT4: Alright. Since Homebrew/homebrew do the same thing I won't disagree.

DomT4:
Merged! Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. For future reference the message about commit wording in your other PR applies here too, and generally unless there's a super-specific need we prefer versions formulae to be pointed at a branch, say ` pandoc113 ` and then with the version being ` 1.13.1 ` but I'll let it slide this time.

randy3k:
do you want me to change it to `pandoc113`?

DomT4:
Na, It's okay this time. Just a future note.

PR #718, opened 2015-03-16
fix formulae style nits

DomT4:
If you aren't a maintainer, please don't submit style-only PRs like this one: I'm only doing it here because I want to generate bottles and versions formulae aren't likely to get updates anytime soon that would create those bottles.

Getting to deal with style nits at the same time is just useful.

If your name is Mike and you created the test-bot system: If this PR breaks the bot horrendously, I deeply apologise <3.


DomT4:
Darn, It was going so well until an old Perl broke on compiling. Will have to play with that one locally a bit.

DomT4:
Alright, gonna have to do the Perl formulae separately. Their build process is a long one and when piled in with everything else causes the build to topple over the hour maximum limit.

DomT4:
Failing on 10.8 - Potentially needs a patch, or a boneyard: allegro4.rb

PR #496, opened 2014-08-05
Adding a CMake 2.8 formula

cloderic:
This provides a formula to build previous 2.8.X versions of CMake

PR #719, opened 2015-03-16
cmake28: 2.8.12.2 - new formula

Tuxity:
Add a older version of cmake. Some project can't work with newer version of cmake, so it could nice to add few older versions, like cmake 3.0 and 3.1 too.

The original formula on main repo is here https://github.com/Homebrew/homebrew/blob/716d4a9b75f7a3e28c7f5d15c5ab5f3beffe1e3b/Library/Formula/cmake.rb


Tuxity:
Just saw an older PR https://github.com/Homebrew/homebrew-versions/pull/496 :x

cmake28.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  satisfy :build_env => false do
+    !File.exist? expat_framework
+  end
+
+  def message; <<-EOS.undent
+    Detected #{expat_framework}
+
+    This will be picked up by CMake"s build system and likely cause the
```
DomT4: I'm afraid find and replace has got some revenge:

> CMake"s build system

:wink:

cmake28.rb
```diff
@@ -0,0 +1,49 @@
...
+    EOS
+  end
+end
+
+class Cmake28 < Formula
+  homepage "http://www.cmake.org/"
+  url "http://www.cmake.org/files/v2.8/cmake-2.8.12.2.tar.gz"
+  sha256 "8c6574e9afabcb9fc66f463bb1f2f051958d86c85c37fccf067eb1a44a120e5e"
+
+  head "http://cmake.org/cmake.git"
```
DomT4: No point having the ` head ` version here. If people want the latest and greatest ` head ` installed, they can get it through Homebrew/homebrew's formula.

DomT4:
Couple of nits, but otherwise this fine :+1:

cmake28.rb
Tuxity: Indeed ahah

cmake28.rb
Tuxity: Oh yeah, I'm fixing it right now

DomT4: Thanks!

Tuxity:
Corrected and squashed

DomT4:
Thanks! Trying to decide if we should make these ` keg_only `, but I think doing so will cause potential compile problems, so I'll probably not bother and just hope nobody tries to install and link 5 versions of CMake alongside each other at once. Thoughts welcome on ` keg_only ` @Homebrew/owners.

mikemcquaid:
If not keg only add `conflicts_with`

DomT4:
Do we want Homebrew/versions ` cmake ` to conflict with each other as well, or just the main ` CMake ` in Homebrew/homebrew?

mikemcquaid:
Each other probably.

DomT4:
WFM.

@Tuxity Can you add a block like this to each of the formulae?

```ruby

conflicts_with "cmake", :because => "both install a cmake binary"
conflicts_with "cmake30", :because => "both install a cmake binary"
```

And so on? It'll just stop people trying to link CMake versions over the top of each other and causing mass-breakage.

Tuxity:
At the top of the file ?

In this case:
```
conflicts_with "cmake", :because => "both install a cmake binary"
conflicts_with "cmake30", :because => "both install a cmake binary"
conflicts_with "cmake31", :because => "both install a cmake binary"
```

And same in other PR as well right ?

DomT4:
Underneath ` depends_on NoExpatFramework ` with a blank line between the blocks. And yeah, in the other PRs as well would be perfect, cheers.

Tuxity:
Done, it's good ?

DomT4:
Yup, looks great, Thanks! :+1:

DomT4:
Rats. Forgot that would happen - Because they're currently being tested individually, the test-bot doesn't know the other CMake versions formulae exist yet. Could you possibly close the other PRs and shunt the two formulae there into this PR as two separate commits? (So commit 1 adds cmake28, Commit 2 adds cmake30 and so on).

Tuxity:
Done !

DomT4:
Looks good so far. The bot is currently having a minor sulk about it's workload, but looks like we should get the green tick soonish. Will merge it in after that.

DomT4:
Merged, Merged and Merged! Thanks for your contribution to Homebrew, and patience in ironing out the few kinks here, we appreciate it! :smiley_cat:. For future reference the preferred commit message for new formulae is ` example12 1.2.3 (new formula) ` - You nearly got it, just needed to remove the ` : ` after the formula name and use ` () ` instead of ` - `.

Tuxity:
Ok! Got it!
Thank you.

PR #720, opened 2015-03-16
cmake30: 3.0.2 - new formula

Tuxity:
Adding cmake 3.0 versions too


cmake30.rb
```diff
@@ -0,0 +1,103 @@
...
+    EOS
+  end
+end
+
+class Cmake30 < Formula
+  homepage "http://www.cmake.org/"
+  url "http://www.cmake.org/files/v3.0/cmake-3.0.2.tar.gz"
+  sha256 "6b4ea61eadbbd9bec0ccb383c29d1f4496eacc121ef7acf37c7a24777805693e"
+
+  head do
```
DomT4: Same thing here, kill the head block.

Tuxity: Should keep the `depends_on "xz" # For LZMA` and put it outside the head block ?

DomT4: Nope, the stable version doesn't make use of ` xz ` here so you can just murder the dep :).

Tuxity:
corrected and squashed :)

Tuxity:
see #719

PR #724, opened 2015-03-18
llvm36: Improve scan-build installation

petere:
Make scan-build-3.6 available from the bin directory in the path by
creating symlinks to the actual installation.  It is necessary to do it
this way because the scan-build program resolves the symlink and finds
its helper files that way.

Related bugs: #558, Homebrew/homebrew#26902, Homebrew/homebrew#33777

### DIFF
--- a/llvm36.rb
+++ b/llvm36.rb
@@ -239,6 +239,10 @@ class Llvm36 < Formula
     system "make", "-C", libcxx_buildpath, "install", *libcxx_make_args
 
     (share/"clang-#{ver}/tools").install Dir["tools/clang/tools/scan-{build,view}"]
+    inreplace share/"clang-#{ver}/tools/scan-build/scan-build", "$RealBin/bin/clang", install_prefix/"bin/clang"
+    ln_s share/"clang-#{ver}/tools/scan-build/scan-build", install_prefix/"bin"
+    ln_s share/"clang-#{ver}/tools/scan-view/scan-view", install_prefix/"bin"
+    (install_prefix/"share/man/man1").install share/"clang-#{ver}/tools/scan-build/scan-build.1"
 
     (lib/"python2.7/site-packages").install "bindings/python/llvm" => "llvm-#{ver}",
                                             clang_buildpath/"bindings/python/clang" => "clang-#{ver}"


llvm36.rb
```diff
@@ -239,6 +239,10 @@ def install
     system "make", "-C", libcxx_buildpath, "install", *libcxx_make_args
 
     (share/"clang-#{ver}/tools").install Dir["tools/clang/tools/scan-{build,view}"]
+    inreplace share/"clang-#{ver}/tools/scan-build/scan-build", "$RealBin/bin/clang", install_prefix/"bin/clang"
```
DomT4: Rather than ` install_prefix/"bin/clang" ` wouldn't just ` bin/"clang" ` work? Same for the other ` install_prefix ` references here as well.

DomT4:
Thoughts welcome from @Homebrew/owners given Adam/Jack/Mike have commented on this issue in the past. Looks good to me though on principle, after taking a glance through the upstream discussions.

mikemcquaid:
I'm OK with stuff like this in the versions formula, just not in core.

llvm36.rb
petere: There is no `bin/"clang"`, only `bin/"clang-x.y"`, which is the same as `install_prefix/"bin/clang"`.  I guess it's a toss-up which one to use, but everything else in the formula operates on `install_prefix`, and then versioned symlinks are made to point to that.

DomT4: Fair enough point. I'm not a particularly huge fan of the ` install_prefix ` language, but it seems to be the accepted norm in the LLVM formulae, which are quite specially written.

DomT4:
Cool. If nobody else in the core team has any objections to this within the next ~24 hours, I'll merge it.

DomT4:
Merged! Slight hiccup with the bottles - Bintray has done something entirely unusual, in having published them but apparently made then inaccessible to anyone who isn't logged in. I'll kill the broken bottles and nudge the bot into creating new ones. Should solve whatever the kink was in the upload process.

Thanks for your contribution to Homebrew anyhow! We appreciate it :smiley_cat:.

DomT4:
Fixed by https://github.com/Homebrew/homebrew-versions/commit/b9ee683b1dc3706c76d4cd16b152bfd811d16703.

PR #735, opened 2015-03-23
Added maven 3.2

rbasoalto:
Maven was recently updated to 3.3.1, breaking compatibility with JDK 1.6. This is a copy of the last version of the maven formula.

PR #736, opened 2015-03-23
Add maven32

mAAdhaTTah:
Just copied 31 -> 32 and changed the link + sha. Should work, but may need to test.


mAAdhaTTah:
whoops, just saw #735.

PR #743, opened 2015-03-31
elasticsearch 1.4.4 (new formula)

ryansch:


elasticsearch14.rb
```diff
@@ -0,0 +1,107 @@
+class Elasticsearch14 < Formula
+  homepage "http://www.elastic.co"
+  url "https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-1.4.4.tar.gz"
+  sha1 "963415a9114ecf0b7dd1ae43a316e339534b8f31"
```
DomT4: This needs to be SHA256.

elasticsearch14.rb
```diff
@@ -0,0 +1,107 @@
+class Elasticsearch14 < Formula
+  homepage "http://www.elastic.co"
```
DomT4: ` https:// ` here.

DomT4:
Couple of nits. Can you integrate [this commit](https://github.com/Homebrew/homebrew/commit/f00dd6af608dd2c928ae23bda54583fb0eda75e9) as well so we fix the problem with config files wandering off with updates? Otherwise, looks great, thanks!

ryansch:
@DomT4 I didn't change the signature from sha1 as that's what the elastic guys publish on the download page.

ryansch:
Rebased into one commit.

DomT4:
@ryansch Please do. SHA256 isn't optional for new formulae - regardless of what upstream do. Plenty of upstream's still display MD5, after all. SHA1 is being phased out by the world fairly rapidly; Microsoft has an open advisory for people to stop relying on it for signing and verification by the end of 2016, and Google has a similar timetable more broadly applied to the web. NIST has also a formal advisory telling people to drop SHA1 by the end of, err... 2013.

We're actively trying to nudge upstream's into displaying SHA256 alongside their favourite checksum, given the chore of doing so is absolutely minimal.

ryansch:
I completely agree that sha1 is busted.  I would be fine with computing the sha256 as long as we include a comment that it doesn't provide any addition protection as I could have gotten a malicious download with a hash collision.  I'll see about poking ES to publish stronger hashes.

DomT4:
The test-bot will test your SHAsum three times - It's fairly unlikely that you would be compromised, and then the three test-bots would all be compromises simultaneously.

ryansch:
I agree that it's unlikely.  I don't want to silently upgrade a user's trust in the transaction when it's really only as good as the sha1 guarantee.  I've opened a ticket with elasticsearch.

DomT4:
Debian, MacPorts, Ubuntu all play the same game - Ignore upstream's checksum and store MD5, SHA1 and SHA256 in each package and then calculate the strongest checksum which the machine supports. We aren't paving new ground here, we've just moving onto the ground that already exists in package management. SHA256 *is* better than SHA1. To screw the checksum you would have to bork the user, bork the test-bots, bork the bottles, bork SHA256 and bork SSL/TLS. SHA1 has likely provable attacks, if you've a spare $700,000. SHA-256 to date does not have a *practical* provable attack.

PR #746, opened 2015-04-02
Add Redis 2.8.19 to versions

leepa:
Unlike the other formulas - we go ahead and alter the binary names so
that you can side-by-side with Redis 3.x (when that gets merged).

We don't alter the port number however so running both at the same time
won't work out of the box - but that's probably a good idea.

### DIFF
--- a/redis28.rb
+++ b/redis28.rb
@@ -1,0 +1,65 @@
+class Redis28 < Formula
+  homepage "http://redis.io/"
+  url "http://download.redis.io/releases/redis-2.8.19.tar.gz"
+  sha256 "29bb08abfc3d392b2f0c3e7f48ec46dd09ab1023f9a5575fc2a93546f4ca5145"
+
+  fails_with :llvm do
+    build 2334
+    cause "Fails with 'reference out of range from _linenoise'"
+  end
+
+  def install
+    # Architecture isn't detected correctly on 32bit Snow Leopard without help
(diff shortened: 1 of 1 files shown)


leepa:
Updated to sha256 to make the bot happy.

mikemcquaid:
@DomT4 Will let you sign this off but looks good to me :+1:

DomT4:
>  we go ahead and alter the binary names so that you can side-by-side with Redis 3.x

Nice! Always a relief to be able to avoid ` keg_only ` or ` conflicts_with ` stuff where possible.

> Will let you sign this off but looks good to me

Thanks! LGTM as well. Downstream bottle build succeeded so I'll merge this in within the next 15 minutes or so :+1:.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/d0dad0e8564ec949b51048feb18ab17b0c292328. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. For future reference please note the preferred commit message for new formulae is ` example12 1.2.3 (new formula) ` - I fixed it up before merge here. Thanks!

joelpresence:
NOTE: there seems to be a bug here.  This does not install redis28.conf or redis.conf.  I have verified this twice.  Please see https://github.com/Homebrew/homebrew-versions/issues/787.  Any chance anybody can fix this?

Thanks!

DomT4:
Fixed by https://github.com/Homebrew/homebrew-versions/commit/ad5a6ee94f34e05e9f17e1cc90ff157e62392ce0.

PR #749, opened 2015-04-03
node08/node010: force npm bump for security fixes

DomT4:
2.7.5 contains two security fixes and consequently falls into the small bracket of reasons as to why the ` node ` formulae would face revision.

Per what I said in the last PR similar to this one, the only reason Node formulae in Versions will get force-bumped outside of normal versioned releases is for security issues. This includes vulnerabilities disclosed by Joyent, the ` npm ` team, or unpatched OpenSSL vulnerabilities existing in older Node versions.

` node08 ` gets a bump to an unreleased ` npm ` - Although it won't be unreleased for long, because ` 2.7.5 ` contained fatal compile errors for the 0.8 branch of Node. Again, this won't happen often. Fairly unique set of circumstances, security fix *and* a compile error.


DomT4:
The bot is blowing itself up. Install is going fine. If this last rebuild doesn't work, I'll merge it anyway. No point retesting endlessly for unrelated ` doctor ` failures.

PR #752, opened 2015-04-07
jpeg formulae: add tests

DomT4:
Spin off of #747 to try and narrow down bottling issues.


DomT4:
@BrewTestBot why do you hate me.

PR #753, opened 2015-04-07
ledger26: improve readability

DomT4:
Closes #747.


DomT4:
Alright. ` ledger26 ` wasn't the problematic formulae of the bigger PR. One down, 5 to go.

PR #755, opened 2015-04-08
povray: fix dead url

### LINKED ISSUE #754: Povray can't be installed
```
$ brew install homebrew/versions/povray36
==> Installing povray36 from homebrew/homebrew-versions
==> Downloading http://www.povray.org/ftp/pub/povray/Official/Unix/povray-3.6.1.tar.bz2

curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "povray36"
Download failed: http://www.povray.org/ftp/pub/povray/Official/Unix/povray-3.6.1.tar.bz2
```

DomT4:
Closes #754


DomT4:
@mikemcquaid Can you look this over? I've resorted to using a method that we don't use in any other formula and the audit is going to throw a tantrum about the unusual style. The ` audit ` be right, but I don't know of a better method here.

We could just plain ` conflicts_with "libpng" ` but it's not strictly true. If you're pouring a bottle, there is no conflict. Conflict only occurs when you have ` libpng ` installed *and* are building from source.

DomT4:
Also a fun and completely unreproducible compile error there. Perhaps one of the ` optional ` deps isn't quite so optional.

mikemcquaid:
Maybe just worth adding a `libpng` dependency?

DomT4:
It breaks horribly against our linpng. Could always put their internal version in versions as a keg_only formula and use that, I guess?

povray36.rb
```diff
@@ -1,31 +1,39 @@
...
-    build 2326
-    cause "povray fails with 'terminate called after throwing an instance of int'"
-  end if MacOS.version == :leopard
+  # Even with the below patch it's still finding libpng when installed.
+  # We can handle this on bottled installs, but not from-source.
+  # Unlinking it forces povray to use its internal libpng over ours.
+  if ARGV.build_from_source?
+    conflicts_with "libpng",
+       :because => "Please unlink libpng during compile. You should link it again after."
+  end
```
mikemcquaid: Maybe just use `build.build_from_source?` in `install`?

mikemcquaid:
Ah, I see. The compilation is failing also which seems unrelated to this? Does this affect the older version? If not, might be best sticking on that.

povray36.rb
DomT4: Didn't think of that. I tried ` if build_from_source? ` but didn't twig it'd likely need an additional ` build `.

> in install

You mean like throw an `onoe ` or something instead of a firm `conflicts_with?`. Probably misunderstanding you there, Sorry :sweat_smile:

DomT4:
> The compilation is failing also which seems unrelated to this?

It looks like one of the deps might not be as optional as the formula hints. It compiled fine locally when I have both deps installed.

> Does this affect the older version? 

Despite the new URL, this wasn't a version bump. Same version, just a new URL. Upstream moved it to an archived part of the website, which is why the new URL looks odd.

mikemcquaid:
Ah, gotcha.

povray36.rb
DomT4: They provide their version of ` libpng ` in their tarball. I think I'm going to piggyback on that to compile their version of ` libpng ` into ` libexec ` and then prepend the ` $PATH `, ` CFLAGS ` and so on. That way we don't have to ` conflicts_with `, we don't have to add new formulae, etc. It should *"just work"*.

Gonna give that a crack.

DomT4: Doesn't work, alas. Still finds our ` libpng ` when linked and goes for that instead, even with the ` $PATH ` and such prepended. :cry:

DomT4:
Just changed it to a basic ` conflicts_with `, with a notice users can and should relink ` libpng ` afterwards again. Meh.

PR #765, opened 2015-04-13
apache spark 1.2.1

jsnrth:
Adds Apache Spark version 1.2.1. The latest version is 1.3.0.

-j


DomT4:
You'll need to fix the errors coughed up by the test bot. ` brew audit --strict ` will also show those errors locally.

jsnrth:
Thanks, I think I got them all, so we'll wait for CI to confirm.  -j

apache-spark121.rb
```diff
@@ -0,0 +1,21 @@
+require "formula"
```
DomT4: Missed this line. Can be killed.

apache-spark121.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class ApacheSpark121 < Formula
+  homepage "https://spark.apache.org/"
+  head "https://github.com/apache/spark.git"
+  url "https://d3kbcqa49mib13.cloudfront.net/spark-1.2.1-bin-hadoop2.4.tgz"
+  version "1.2.1"
+  sha256 "8e618cf67b3090acf87119a96e5e2e20e51f6266c44468844c185122b492b454"
+
+  conflicts_with "hive", :because => "both install `beeline` binaries"
```
DomT4: ` conflicts_with "apache-spark" ` as well, for obvious reasons.

jsnrth: Ah, right. No, they're not, it will conflict.

jsnrth:
I'm confused about the CI error. It seems to be mavericks only?  -j

apache-spark121.rb
```diff
@@ -0,0 +1,20 @@
...
+  conflicts_with "apache-spark"
```
DomT4: Can you add a note here as well? ` :because => "Both install the same binaries" ` is fine. Otherwise we'll get people asking, thanks!

DomT4:
Mavericks crash wasn't your fault. Bot threw a fatal but unrelated error. Once you've added the comment the bot will look over the PR again.

apache-spark121.rb
jsnrth: Ok, will do.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/7d1f5add1ab2f620d7bf6167d895cba6b06a7f0f. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. For future reference the preferred commit message for new formulae is ` apache-spark121 1.2.1 (new formula) `, Thanks!

jsnrth:
Great, thanks! -j

PR #766, opened 2015-04-13
GnuPG2.1: add option to install gpgsplit as well

niobos:
gpgsplit is part of the gnupg package, but is not installed by default. This adds a --with-gpgsplit option to request the installation.

An alternative would be to split this in to its own package/formula.


gnupg21.rb
```diff
@@ -11,6 +11,8 @@ class Gnupg21 < Formula
     sha256 "36a4821c7a05225b1b3c57b618ed37c4194e46615f76054b9dcb628405b0fe96" => :mountain_lion
   end
 
+  option 'with-gpgsplit', 'also install gpgsplit'
```
DomT4: Double quotes everywhere. Suggest the message be a little more formal - ` "Additionally Install the 'gpgsplit' utility" ` or similar

gnupg21.rb
```diff
@@ -91,6 +93,8 @@ def install
     system "make"
     system "make", "check"
     system "make", "install"
+    system "make", "-C", "tools", "gpgsplit" if build.with? 'gpgsplit'
```
DomT4: Double quotes again at the end there.

DomT4:
Why isn't this building automatically anymore? It did for the 1.x and 2.0.x branches.

gnupg21.rb
```diff
@@ -91,6 +93,8 @@ def install
...
+    bin.install "tools/gpgsplit"
```
DomT4: We'll have to rename this, since it conflicts with the ` gnupg ` 1.x branch.

Something like this is fine => ` bin.install "tools/gpgsplit" => "gpgsplit2" `.

niobos:
I've changed the quotes, and renamed the binary.
I've also renamed the option to `--with-tools`, to match what `gnupg.rb` calls it.

I'll look in to the automatic building later.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/b330bd8e59ccdcf0f350ff553dad7e828ef3b58a. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. For future reference the ideal commit message here would have been ` gnupg21: add option to install gpgsplit `. Lowercase rather than uppercase, essentially. Thanks!

PR #769, opened 2015-04-13
gnupg21: fix clang & llvm build

DomT4:
Upstream committed a patch to fix Clang & LLVM compile. Thanks Werner.


DomT4:
Failure unrelated. Merging soon.

PR #773, opened 2015-04-14
libmongoclient-legacy: add 1.0.1 legacy branch (new formula).

billybones256:
libmongoclient-legacy: stable legacy branch of the official mongo-cxx-driver (new formula).


billybones256:
Added a `conflicts_with` to avoid bumping into Homebrew/homebrew's Libmongoclient formula.

libmongoclient-legacy.rb
```diff
@@ -0,0 +1,45 @@
...
+      # when using libc++
+      "--osx-version-min=#{MacOS.version}",
+      "install",
+    ]
+
+    args << "--libc++" if MacOS.version >= :mavericks
+
+    scons *args
+  end
+  test do
```
DomT4: Can this test be modified to do something more substantial than e.g. --version or --help (e.g. create a anonymous gist)? See cmake.rb for an example of an application formula with a good test and tinyxml2.rb for an example of a library formula with a good test. Thanks!

libmongoclient-legacy.rb
```diff
@@ -0,0 +1,45 @@
+class LibmongoclientLegacy < Formula
+  homepage "http://www.mongodb.org"
```
DomT4: ` https:// ` rather than ` http:// `.

billybones256: Done.

libmongoclient-legacy.rb
billybones256: Removed. This option is necessary only for 26compat branch.

DomT4: Sorry, I didn't mean the ` libc++ ` part, I meant can the ` test do ` stage be improved?

billybones256: It’s possible to improve test do stage only if the user has installed mongodb on localhost because tests depends on it. I included do stage only because of the recommendations in the Formula-Cookbook.
> 15 апр. 2015 г., в 21:11, Dominyk Tiller <notifications@github.com> написал(а):
> 
> Sorry, I didn't mean the libc++ part, I meant can the test do stage be improved?

DomT4: Ah. In which case, just remove the ` test do ` block completely. It'll cause the bot to red-flag the lack of test, but I can ignore that.

billybones256: Ok, I did it.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/02620488304012c50203192faf962f0a5c1f4210. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

For future reference please note the commit message style here ideally would have been ` libmongoclient-legacy 1.0.1 (new formula) ` and squashed commits = happy maintainers :wink:. Thanks!

PR #774, opened 2015-04-15
postgis20 2.0.7

### LINKED ISSUE #772: Failed to install posgis20
Hi
I spent 4 hours trying to install posgis on a postgres 9.2 installation (done with homebrew) but 
can'f figure how to solve my issue.
I am on Yosemite and use the last version of homebrew
```
brew install -vd postgis20
...
configure: error: Cannot find json dev files in "/usr/local/opt/json-c"
/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: ./configure --disable-dependency-tracking --with-projdir=/usr/local --with-jsondir=/usr/local/opt/json-c --with-pgconfig=/usr/local/Cellar/postgresql92/9.2.8_1/bin/pg_config --disable-nls

but json-c is installed properly:
xxxxxx@xxxxxx /usr/lib $ ls -all /usr/local/opt/json-c/lib/
total 352
drwxr-xr-x 6 xxxxxx admin 204B 13 Apr 16:55 ./
drwxr-xr-x 11 xxxxxx admin 374B 13 Apr 16:55 ../
-r--r--r-- 1 xxxxxx admin 76K 13 Apr 16:55 libjson-c.2.dylib
-r--r--r-- 1 xxxxxx admin 90K 13 Apr 16:55 libjson-c.a
lrwxr-xr-x 1 xxxxxx admin 17B 13 Apr 16:55 libjson-c.dylib@ -> libjson-c.2.dylib
drwxr-xr-x 3 xxxxxx admin 102B 13 Apr 16:55 pkgconfig/
```
Can someone help me?

DomT4:
Version bump. Fixes recent json-c versions detection.

Closes #772.


DomT4:
@Homebrew/owners @xu-cheng Found the first formula which breaks under sandboxing. We're going to need to exempt the Postgis formulae, or stop supporting them entirely. Their install method gets butchered by the sandboxing when it tries to start placing its built files in other formulae directories. Throws ` Operation not permitted ` messages everywhere.

xu-cheng:
Can we work around using `post_install`? Although post install sandbox hasn't been merged yet, it will allow to write entire HOMEBREW_PREFIX.

DomT4:
Does ` post_install ` have access to install's ` buildpath `? Otherwise I guess we could dump everything in ` libexec ` during install and then move it all around ` postinstall `. If we ever did manage to bottle ` postgis `' family of formulae we'd *have* to do the ` libexec ` thing since we don't bottle the ` buildpath `.

Presumably we'd have to symlink though in ` postinstall ` otherwise users running `brew post-install postgis ` twice in a row would quickly hit failure as the files it would have installed elsewhere. Likewise, I don't think we currently have a solution in place to automatically remove files juggled around in ` postinstall ` so they'd remain on the system.

It's a complicated solution if we go down the ` postinstall ` route. It's probably one of the better solutions we have, but I think beating the Postgis formulae into the sandbox is going to be a fun exercise. One I'm in favour of, but fun.

mikemcquaid:
We should fix it so it doesn't need to write into other Cellars. If we can't do that we'll need to use either `post_install` or we remove it.

DomT4:
Looking at the notes in-formula I'm not sure how possible the former is. I suppose it may be possible for us to merge this with the appropriate ` postgres ` formulae and vendor it as a resource, since it only links into its own prefix and Postgres' prefix at the moment. That would make the Postgres formulae behemoths though, and not terribly pleasant for anyone to maintain. `post_install` looks like the best bet for the future, if we don't mind things not being deleted if the `postgis` formulae are deleted.

xu-cheng:
Now since you are mentioning, merging it into Postgres as a build option is a better choice. At the same time, it fits the policy.

DomT4:
I'm more comfortable with the `post_install` option personally, if only because Postgres is already a pain in the neck to maintain and fairly frequently is the cause of issues at Homebrew/homebrew. Kinda worried pouring another formula on top of that is going to be painful. Happy to be out-voted though.

Will merge this in shortly, to resolve the immediate build error we have here. Can continue to work out the long-term Postgis issues, but I don't want to leave the formula broken until we do, heh.

PR #778, opened 2015-04-17
Include flag --with-build-config=bootstrap-debug

leonardoobaptistaa:
Build with Xcode 6.3 is broken. This PR is just a update from the original repo.
Original PR: https://github.com/Homebrew/homebrew/pull/38596
Explanation: https://github.com/mriddle/puppet-gcc/releases/tag/2.2.1

PR #795, opened 2015-04-25
gcc48: add test

DomT4:
Catch up with audit changes upstream, change to SHA256, add test, change the options to match every other gcc formulae we ship, etc.


DomT4:
@Homebrew/owners Any objections/consequences to scoping the ` "--with-build-config=bootstrap-debug" ` Yosemite build fix exclusively to Yosemite here? Tried adding an ` LDFLAG ` in #778 but didn't seem to make much difference, but without the ` bootstrap-debug ` fix in the ` gcc48 ` formula the formula doesn't compile on 10.10 at all. I could just be being an idiot and suggesting the wrong ` LDFLAG ` - Feel free to call me an idiot if I'm being one :wink:.

mikemcquaid:
@DomT4 I don't mind, seems fine to me.

DomT4:
Alright, pushed. If it passes the bot and nobody else throws an objection, I'll merge this sometime tomorrow.

DomT4:
Merged.

PR #796, opened 2015-04-25
nu0: fix dead url

DomT4:
The URL is dead and I can't find a mirror for it anywhere. Unless anyone has one, it'll have to be boneyarded.


DomT4:
CC @Homebrew/owners

mistydemeo:
Macports has it in their distfiles repository: http://distfiles.macports.org/nu/Nu-0.4.0.tgz

DomT4:
Nice! I checked Debian but forgot MacPorts mirror a decent amount of things as well. Thanks Misty, appreciate that.

DomT4:
This is enthusiastically refusing to build for me locally. Doesn't even make it far it before it blows up. I'm going to let the bot run over it to see if it compiles on any of the three OS X versions bottles support, but otherwise we may want to MMOS Req it perhaps?

PR #800, opened 2015-04-27
Add the --disable-warnings-as-errors flag for building Mongo 2.6

dannyfallon:
The latest versions of Xcode on Yosemite come with a version of
clang that will throw errors when compiling 2.6 from source. Since
MongoDB are unwilling to compensate with backporting the fixes from
2.7/3.0 to 2.6 we'll use their build flag. Without this it's
currently impossible to install from source on a fully up to date
Yosemite.


dannyfallon:
The PR I sent to Mongo where they told me to use the build flag: https://github.com/mongodb/mongo/pull/956

I'd really like at some stage to get clang version detection into their Scons build file, I think tha

mongodb26.rb
```diff
@@ -41,6 +41,11 @@ def install
     args << "--use-system-boost" if build.with? "boost"
     args << "--64" if MacOS.prefer_64_bit?
 
+    # Pass the --disable-warnings-as-errors flag to Scons when on Yosemite
+    # or 2.6.x won't build from source due to a Clang 3.5+ error
+    # https://github.com/mongodb/mongo/pull/956#issuecomment-94545753
+    args << "--disable-warnings-as-errors" if MacOS.version == :yosemite
```
DomT4: If this is going to impact future OS X versions as well, this should be ` if MacOS.version >= :yosemite `.

DomT4:
One little nit, but aside that given this is upstream's advice I'm happy enough to carry it.

dannyfallon:
You're right @DomT4, I've rebased so we're now using >= as the comparison operator :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/44a17aab61cda9153c99987eff4a76121c1f1622. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just for future reference, the preferred commit message for formula commits that aren't version bumps is ` mongodb26: add --disable-warnings-as-errors flag for building ` and similar. Thanks!

dannyfallon:
Thanks, I'll be sure to follow the commit message guidelines in future :ok_hand:

PR #805, opened 2015-04-29
elasticsearch: changed domains for all formulae

sj26:
Elasticsearch have changed domain names. The old download URLs no longer work. Changing the host inside the URLs seems to make them work again.

### DIFF
--- a/elasticsearch-0.20.rb
+++ b/elasticsearch-0.20.rb
@@ -3,7 +3,7 @@ require 'formula'
 class Elasticsearch020 < Formula
   homepage 'https://www.elastic.co/products/elasticsearch'
   url 'https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-0.20.6.tar.gz'
-  sha1 'f66a778ad94ea1dd69d18f8f89ce32c2383898eb'
+  sha256 '534b9fdb2fa9031f539b19cc4cfc1345bbbb013a285a21105ca87348b96b3389'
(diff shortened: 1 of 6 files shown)


elasticsearch-0.20.rb
```diff
@@ -1,8 +1,8 @@
 require 'formula'
 
 class Elasticsearch020 < Formula
-  homepage 'http://www.elasticsearch.org'
-  url 'http://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.20.6.tar.gz'
+  homepage 'https://www.elastic.co/products/elasticsearch'
+  url 'http://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-0.20.6.tar.gz'
```
DomT4: Can you bump this ` url ` to ` https:// ` as well please? Thanks!

DomT4:
Thanks for the PR! Looks good. One nit with the ` https:// ` link and I'm cool with this. If you wanted to change all the SHA1 sums in the ` url ` block to `sha256` I certainly wouldn't object, but it's not mandatory.

sj26:
Updated! :-)

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/b4137aabb62c662c48823b68d218fe34eb54454a. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #809, opened 2015-05-02
mono3 (new formula)

xu-cheng:
Ref: https://github.com/Homebrew/homebrew/issues/39127


mono3.rb
```diff
@@ -0,0 +1,76 @@
...
+        </ItemGroup>
+        <Import Project="$(MSBuildBinPath)\\Microsoft.CSharp.targets" />
+      </Project>
+    EOS
+    shell_output "#{bin}/xbuild test.csproj"
+  end
+
+  def caveats; <<-EOS.undent
+    To use the assemblies from other formulae you need to set:
+      export MONO_GAC_PREFIX="#{HOMEBREW_PREFIX}"
```
DomT4: Is this still true if this Mono is `keg_only`?

xu-cheng: I'm not a mono user, so honestly I don't know.

xu-cheng: ping @directhex for help.

directhex: I'm not familiar with the terminology. What does "keg_only" mean?

xu-cheng: It means this mono installations wouldn't be symlinked to `HOMEBREW_PREFIX` in order to avoid the conflicts with mono formula in the core.

DomT4: Maybe ` conflicts_with ` is a better fit here?

xu-cheng: I think `keg_only` is fine here. And this caveats seem only related to what mono will read. So Let's merge this as it is? Unless I'm wrong.

directhex: There's definitely a conflict with the symlinking - the commands aren't versioned, so keg_only seems fine to me

DomT4: *Shrug* No idea. If nobody else files any objection to boxing it into its prefix I'll merge in a couple hours.

DomT4: (Or you can merge it of course. No objections from me if you want to).

PR #816, opened 2015-05-07
play12 1.2.7.2

dunn:
- couldn't get the test to work with a plain `IO.popen`
- should the executable be renamed `play12` rather than introducing a conflict?

### DIFF
--- a/play12.rb
+++ b/play12.rb
@@ -1,15 +1,26 @@
-require 'formula'
-
 class Play12 < Formula
-  homepage 'http://www.playframework.org/'
-  url 'http://downloads.typesafe.com/play/1.2.7/play-1.2.7.zip'
-  sha1 '436739d9f7cc00567a7e4245413c9c1ebf886797'
+  homepage "http://www.playframework.org/"
+  url "http://downloads.typesafe.com/play/1.2.7.2/play-1.2.7.2.zip"
+  sha256 "4fc610e2db0993fee3daf01527ba5ca57a9ad970c429733415ee59b55064b322"
+
+  conflicts_with "sox", :because => "Both install a `play` executable"
 
   def install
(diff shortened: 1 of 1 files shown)


play12.rb
```diff
@@ -1,15 +1,26 @@
...
+
+  conflicts_with "sox", :because => "Both install a `play` executable"
 
   def install
-    rm_rf 'python' # we don't need the bundled Python for windows
-    rm Dir['*.bat']
-    libexec.install Dir['*']
-    bin.mkpath
-    ln_s libexec+'play', bin
+    rm_rf "python" # we don"t need the bundled Python for windows
```
DomT4: Replace all has bitten you here :)

DomT4:
> should the executable be renamed play12 rather than introducing a conflict?

We usually just conflict. Generally don't rename binaries unless it's something users expect, i.e. it happens upstream and/or other package managers. Open to persuasion though if it happens elsewhere.

play12.rb
xu-cheng: @dunn A small tip:

Instead of using global replace to fix double quotes, you can just run `brew style --fix <formula>` if `HOMEBREW_DEVELOPER` is set.

dunn: @xu-cheng that's great to know, thanks!

dunn:
We don't add `conflicts_with` to core formula when the conflict is with a tap, right? `sox.rb` doesn't need to be changed?

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/bf77f34ee6283969ba555315a3c66f2c4715de19, Thanks!

> We don't add conflicts_with to core formula when the conflict is with a tap, right?

Nope, core formulae remain untouched in this situation :+1:.

PR #819, opened 2015-05-09
play13 1.3.1

dunn:


play13.rb
```diff
@@ -10,12 +10,26 @@ class Play13 < Formula
     sha1 "260e3924f28c9f7df7b54a7e90d4a03a764691e7" => :mountain_lion
   end
 
+  conflicts_with "sox", :because => "Both install a `play` executable"
+  conflicts_with "play12", :because => "Both install a `play` executable"
```
dunn: Does this need the full name `homebrew/versions/play12`?

DomT4: Since it's in the same tap, nope :).

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/9fe827d136967bc3230a228e039da876fc7c70ca, Thanks again! :heart:.

PR #827, opened 2015-05-13
protobuf240a: fix dead url

DomT4:
Upstream moved to Github, didn't take this download with them, Debian/Ubuntu/MacPorts don't mirror.


DomT4:
CC @Homebrew/owners for agreement/disagreement.

mikemcquaid:
I'd consider Googling for `protobuf-2.4.0a.tar.bz2` and finding something with a matching checksum. Personally I'm of the view that we should be less aggressive boneyarding stuff from versions.

DomT4:
I did a fairly thorough 8+ page Google and DuckDuckGo hunt before I opened this PR. I also checked Debian, Ubuntu and MacPorts for mirrors.

Somewhat miraculously, Launchpad has a working copy even though it's not in the main Linux mirrors. It's `tar.gz` rather than what we ship now, so obviously the checksum can't be verified at all, but apparently Launchpad are the only people in the world still mirroring it.

Guess it gets to live to fight another day. Presuming it still compiles.

mikemcquaid:
I dunno, if it's `.tar.bz2` and in a place on LaunchPad that could go away in future that might be reason to boneyard it.

DomT4:
It hasn't moved from Launchpad in 3 years. Unless we get super unlucky I can't imagine it'll vanish soon. I managed to verify the checksum of the download against Debian's release from the time. If you want to keep carrying it, I'm happy enough on the checksum verification to change the PR to a url change.

mikemcquaid:
Cool, works for me :+1:

PR #831, opened 2015-05-16
gcc5: disable Go support

### LINKED ISSUE #830: Does gcc5's gccgo actually work?
As of gcc 4.9 it didn't. If it doesn't work, we should remove the option.

rhencke:
Closes #830.

### DIFF
--- a/gcc5.rb
+++ b/gcc5.rb
@@ -33,9 +33,10 @@ class Gcc5 < Formula
     sha256 "5cc7eaaab1dd29879109ccf896f91c691ba6267a997288ee526e3f21479a8f02" => :mountain_lion
   end
 
+  # GCC's Go compiler is not currently supported on Mac OS X.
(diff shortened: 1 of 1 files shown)


rhencke:
(Sorry, I think I confused the build bot by pushing a new commit to this request mid-build.  Is there a way to re-trigger the build?)

DomT4:
Don't worry about it. We're not updating the bottle since it doesn't change anything that would or wouldn't already be in the bottle, so we can just pull the change.

It passed the one bot which shows you didn't cause any audit failures, and since the `go` build element was never tested in the first place because of it being `optional` I'm happy to pull without an across-the-board green tick.

rhencke:
Much obliged - thank you. :)

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/24ef9ca402cbb8db88b6b78f7b9bec3fc91c9154. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #832, opened 2015-05-17
Pullrequests/solr510

Philippe-Caparroy:


DomT4:
Commits need squashing here as well please. I'm not against it, but curious why you're submitting it to versions when this very version already exists in the Core? We do this for other things, but it's not super-common enough that my curiosity isn't piqued.

Philippe-Caparroy:
Ok, I'm sorry. Firstly I didn't noticed that the solr formula in the core was just for the latest version. So there are no reasons to re submit to homebrew-versions. Should I do something to close my pull request?
 Secondly I apparently misunderstood that the core was for latest versions? Therefore should I submit the ganglia formula to the core, as the ganglia formula allready present in the core is 3.6.0, and the one I submitted is for the latest release : 3.7.1?
Thanks in advance.

DomT4:
Yeah. Homebrew/homebrew always takes the latest stable version of something, *unless* that new stable version is drastically incompatible with an existing dependency, dependant, or is just likely to break things for users more regularly than expected. Versions only takes newer stuff when the above is true, generally.

Unless one of the above is true, you should submit the `ganglia` PR in Versions here as an update to the existing `ganglia` in Homebrew/homebrew :).

DomT4:
Closing here. Will welcome resubmission once the Homebrew/homebrew formula is no longer on `solr-5.1.0`, Thanks!

PR #837, opened 2015-05-23
Rubies

DomT4:
Prune dead mirrors, fix OpenSSL linkage, add tests, fix style issues, etc.


DomT4:
Sigh. Mountain Lion. Looks like a race, will kill the parallel and try again.

PR #839, opened 2015-05-24
libmongoclient-legacy: add 1.0.2

billybones256:
Update mongodb cpp driver to 1.0.2


DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/a161983cb9d9af8118f3fe49dd0bee0dd351e0fa. Thanks for your contribution to Homebrew, we appreciate it :smiley_cat:.

Couple of notes for future: Please leave the bottle block intact, we replace it automatically on update but it can upset the logic if it's removed and readded in the same "event", and the ideal commit message for simple version bumps is `libmongoclient-legacy 1.0.2`, Thanks!

billybones256:
Ок.

PR #847, opened 2015-06-01
Remove allegro5.

SiegeLord:
Moving allegro5 to the core tap, see: https://github.com/Homebrew/homebrew/pull/40258


DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/269dcbe12c51e8537e6899314b18d6dbb6396489. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a wee note: For migration from taps to core the commit message should be something like `example: migrate to homebrew/example` - That way it's reasonably obvious where the formula has gone.

PR #848, opened 2015-06-02
Create libxml278

rogeriopradoj:
Very useful for compiling PHP 5.2 and other old stuff, as pointed in:
http://www.devsumo.com/technotes/2014/06/php-node-lo-compiler-errors-when-building-on-os-x-10-9/


rogeriopradoj:
I've seen the build failed, I'll work on it.

rogeriopradoj:
Done!

libxml278.rb
```diff
@@ -0,0 +1,38 @@
+class Libxml278 < Formula
+  url "ftp://xmlsoft.org/libxml2/libxml2-2.7.8.tar.gz"
+  homepage "http://xmlsoft.org"
+  sha256 "cda23bc9ebd26474ca8f3d67e7d1c4a1f1e7106364b690d822e009fdc3c417ec"
+  desc "Libxml2 is the XML C parser and toolkit"
```
DomT4: This is slightly in the wrong order. Should be `desc` => `homepage` => `url` => `sha256`.

libxml278.rb
```diff
@@ -0,0 +1,38 @@
...
+    ENV.universal_binary if build.universal?
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}",
+                          "--without-python"
+    system "make"
+    ENV.j1
+    system "make", "install"
+
+    if build.with? "python"
```
DomT4: There is no `depends_on :python => :optional`, so this will never be triggered. Either remove it or add the optional dep :)

libxml278.rb
rogeriopradoj: Done!

libxml278.rb
rogeriopradoj: Done!

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/682f1fbdb06501eb1fd89059db61a7839f9c27ba. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note the preferred commit message style for new formulae is `libxml278 2.7.8 (new formula)`, Thanks!

PR #859, opened 2015-06-10
mysql55: updates from core

tamird:
This was pretty frustrating to debug!


mysql55.rb
```diff
@@ -106,6 +106,7 @@ def install
     ln_s "#{prefix}/support-files/mysql.server", bin
 
     # Move mysqlaccess to libexec
+    libexec.mkpath
```
DomT4: There's a lot of syntax in this formula we've stopped using or improved since. The direct calls to `ln_s` and `mv` in particular.

Here, we could be able to do this instead of adding another line to create `libexec`:

```ruby
libexec.install bin/"mysqlaccess"
libexec.install bin/"mysqlaccess.conf"
```

That should automatically create the directory and then move the files there.

tamird: heh, and yet: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mysql.rb#L130

DomT4: Heh, a classic case of [committed](https://github.com/Homebrew/homebrew/commit/b8a121e056a754a7b6bab29aaf7624df9bf0e6cb) over a year ago and nobody has glanced at it since. It's fine to change, just checked locally.

Jack fixed the `ln_s` logic in https://github.com/Homebrew/homebrew/commit/34ad88ddf140837dad474399a14a5c61b0ea68da, which is why that's no longer around in the core. You can tweak that here too if you like.

tamird:
Green

DomT4:
https://github.com/Homebrew/homebrew-versions/commit/1dc3065c8800df5346c36a2a6dfb5723078f9c7e :rocket:

PR #867, opened 2015-06-15
Add brew support for nasm version 2.11.06

ArtillerySr:
This code was taken from a previous core nasm build. Please help me fix it up so that is it in line with homebrew-versions standards.

### DIFF
--- a/nasm21106.rb
+++ b/nasm21106.rb
@@ -1,0 +1,28 @@
+class Nasm21106 < Formula
+  desc "Nasm 2.11.06 Formula"
+  homepage "http://www.nasm.us/"
+  url "http://www.nasm.us/pub/nasm/releasebuilds/2.11.06/nasm-2.11.06.tar.xz"
+  sha256 "90f60d95a15b8a54bf34d87b9be53da89ee3d6213ea739fb2305846f4585868a"
+
+  option :universal
+
+  def install
+    ENV.universal_binary if build.universal?
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install", "install_rdf"
+  end
+
+  test do
+    (testpath/"foo.s").write <<-EOS
+      mov eax, 0
+      mov ebx, 0
+      int 0x80
+    EOS
+
+    system "#{bin}/nasm", "foo.s"
+    code = File.open("foo", "rb") { |f| f.read.unpack("C*") }
+    expected = [0x66, 0xb8, 0x00, 0x00, 0x00, 0x00, 0x66, 0xbb,
+                0x00, 0x00, 0x00, 0x00, 0xcd, 0x80]
+    assert_equal expected, code
+  end
+end


nasm21106.rb
```diff
@@ -0,0 +1,29 @@
...
+      int 0x80
+    EOS
+
+    system "#{bin}/nasm", "foo.s"
+    code = File.open("foo", "rb") { |f| f.read.unpack("C*") }
+    expected = [0x66, 0xb8, 0x00, 0x00, 0x00, 0x00, 0x66, 0xbb,
+                0x00, 0x00, 0x00, 0x00, 0xcd, 0x80]
+    assert_equal expected, code
+  end
+end
```
DomT4: Needs a new, blank line a the end.

nasm21106.rb
```diff
@@ -0,0 +1,29 @@
+require "formula"
```
DomT4: Can remove this completely.

nasm21106.rb
```diff
@@ -0,0 +1,29 @@
+require "formula"
+
+class Nasm < Formula
```
DomT4: The test-bot failures are primarily because of the formula name mismatch - The file is called `nasm21106` but inside the Formula it defines its name as just `Nasm`. It should be `class Nasm21106 < Formula`

ArtillerySr:
Not really sure what the build error here is. Looks as if it couldn't locate this formula at all.

DomT4:
It crashed. I've asked it to look over the PR again.

ArtillerySr:
Looks as if we have another crash here.

DomT4:
Can you squash your commits down to 1? I'm wondering if that's the cause here, since nothing else is crashing this regularly.

ArtillerySr:
8 Commits have been squashed. Sorry for the mess.

DomT4:
Thanks! Still blowing up on the bot, which is odd. @jacknagel @mikemcquaid Any idea why this has crashed Jenkins 5 times in a row? I feel like I'm missing something obvious, but :confused:

jacknagel:
Feedback from the 10.10 bot:

```
==> brew style nasm21106
== /usr/local/Library/Taps/homebrew/homebrew-versions/nasm21106.rb ==
C: 26:  3: Inconsistent indentation detected.

1 file inspected, 1 offense detected
```

ArtillerySr:
Cleaned up that last indentation error however the bot still seems to be crashing. I agree with @DomT4 it seems like we are missing something obvious here.

jacknagel:
I don't think there's anything wrong with this PR, the failure is happening before the bot even pulls it. It completed an entire CI run when I started it manually yesterday.

ArtillerySr:
Looks good to go. Thanks for all the help.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/c6de18f84701a4e57f1894a089cf93d89ed8d1c3. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #870, opened 2015-06-16
llvm36: Patch to fix --with-asan

### LINKED ISSUE #756: llvm36 fail to build in 10.10 if add --with-asan
Latest Command Line Tools is installed.

Output of `brew install llvm36 --with-lldb --with-lld --with-all-targets --with-asan`:

<https://gist.github.com/anonymous/b241e76a120ecccd76c3>

    make[3]: *** [compiler-rt/.makeall] Error 2
    make[2]: *** [all] Error 1
    make[1]: *** [clang/.makeall] Error 2
    make: *** [all] Error 1

It fail to build because of option "--with-asan", if i use `brew install llvm36 --with-lldb --with-lld --with-all-targets`, it will success to build.

So where the problem is?

flaub:
Fixes #756.

### DIFF
--- a/llvm36.rb
+++ b/llvm36.rb
@@ -143,6 +143,7 @@ class Llvm36 < Formula
     (buildpath/"tools/lld").install resource("lld") if build.with? "lld"
     (buildpath/"tools/lldb").install resource("lldb") if build.with? "lldb"
     (buildpath/"projects/compiler-rt").install resource("compiler-rt") if build.with? "asan"
+    inreplace "tools/clang/runtime/compiler-rt/Makefile", /CC=\"\$\(ToolDir\)\/clang\" \\/, "\\" if build.with? "asan"
 
     if build.universal?
       ENV.permit_arch_flags


llvm36.rb
```diff
@@ -143,6 +143,7 @@ def install
     (buildpath/"tools/lld").install resource("lld") if build.with? "lld"
     (buildpath/"tools/lldb").install resource("lldb") if build.with? "lldb"
     (buildpath/"projects/compiler-rt").install resource("compiler-rt") if build.with? "asan"
+    inreplace "tools/clang/runtime/compiler-rt/Makefile", /CC=\"\$\(ToolDir\)\/clang\" \\/, "\\" if build.with? "asan"
```
flaub: I don't know if this is the best way to accomplish this. I want to patch the file but wasn't clear about how to use an embedded patch.

flaub:
This can likely be applied to llvm34 and llvm35 as well.

DomT4:
We've reverted various workarounds for `asan` before now, so will let other @Homebrew/owners have some time to review here.

llvm36.rb
DomT4: Can we report the need for doing this upstream? I'd like to keep track on it long-term. Perhaps the LLVM team can flag up whether this is something we're causing or something that's a core LLVM issue.

You can also block the two "asan" calls together, like this:

```ruby
if build.with? "asan"
  (buildpath/"projects/compiler-rt").install resource("compiler-rt")
  # https://bugreport.submitted.com - Fix asan compile
  inreplace "tools/clang/runtime/compiler-rt/Makefile", /CC=\"\$\(ToolDir\)\/clang\" \\/, "\\"
end
```

apjanke:
Ping on this? It's been a year, so I'm inclined to close this unless there's been upstream bug report activity.

PR #873, opened 2015-06-24
Added jython 2.5.3

seletz:
Added new formula for jython version 2.5.3.

### DIFF
--- a/jython253.rb
+++ b/jython253.rb
@@ -1,5 +1,5 @@
 class Jython253 < Formula
-  desc "Python implementation written in Java (successor to JPython)"
+  desc "Python on the JVM"
   homepage "http://www.jython.org"
   url "https://search.maven.org/remotecontent?filepath=org/python/jython-installer/2.5.3/jython-installer-2.5.3.jar"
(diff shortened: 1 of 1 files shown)


seletz:
Ok, seems like I need to do some RTFM.

jython253.rb
```diff
@@ -0,0 +1,11 @@
+class Jython < Formula
```
DomT4: The `class` name here needs to match the formula name, which is the error you're getting.

DomT4:
Couple nits to chase up on:

```
==> audit problems
homebrew/versions/jython253:
 * Description is too long. "name: desc" should be less than 80 characters (currently 89).
```

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/97f9545144fc07637ba268d1bb6c60e6d675b11f. Thank you for your contribution to Versions, we appreciate it! :smiley_cat:.

Just a nit, but new formulae commit messages should be styled `jython253 2.5.3 (new formula)` or similar, Thanks!

PR #874, opened 2015-06-26
Created formula for installing Ant version 1.7.1

arigesher:
Sometimes you just need an ancient version of ant and it would be great to brew install it.  Ant 1.7.x suffered from an API break in the 1.8.x series, so here is the final version of 1.7.x.


DomT4:
You probably need to add a `depends_on :java` dep here to help the `test` find Java properly.

DomT4:
Ping?

DomT4:
Closing as dead. Feel free to resubmit the PR with the necessary fix, Thanks for the submission!

PR #880, opened 2015-07-04
node010/node08: npm bump

DomT4:
The update to npm 2.12.1 should resolve the remaining issues around people accidentally/deliberately running npm with sudo and perma-breaking their installations. `node08` also gets a fix for the manpage symlinking problem that we previously fixed in 0.12.x and 0.10.x Node formulae.


DomT4:
Barring any surprises, this should mark the end of revisioning for the `npm` 2.x.x branch. Hoping with the root problem solved it'll be easy sailing for a fair while.

PR #882, opened 2015-07-06
add python 3.3

aep-shoutlet:
Add formula for Python v3.3.

### DIFF
--- a/python33.rb
+++ b/python33.rb
@@ -1,0 +1,364 @@
+require 'formula'
+
+class Python3 < Formula
+  homepage 'http://www.python.org/'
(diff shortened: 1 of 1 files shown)


DomT4:
There's a lot of `brew audit --strict` stuff here that needs fixing, as well as the `Class` name which should match the formula name. @tdsmith has had some past concerns on accept Python formulae into Versions given the potential for clashing; if Tim's still against it I'm a :-1: on this.

tdsmith:
Pyenv is a great way to use multiple Python versions or you can maintain this yourself in a tap. Thanks!

PR #885, opened 2015-07-07
Add recipe giflib41.rb

karmakaze:
Giflib 4.2 has breaking api changes so make recipe for 4.1 version.

### DIFF
--- a/giflib41.rb
+++ b/giflib41.rb
@@ -1,0 +1,18 @@
+class Giflib41 < Formula
+  desc "A library and utilities for processing GIFs"
+  homepage "http://giflib.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.1.6/giflib-4.1.6.tar.bz2"
+  sha256 "e1c1ced9c5bc8f93ef0faf0a8c7717abf784d10a7b270d2285e8e1f3b93f2bed"
+
+  keg_only "Conflicts with giflib in main repository."
+
+  def install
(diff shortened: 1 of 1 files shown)


giflib41.rb
```diff
@@ -0,0 +1,24 @@
+class Giflib41 < Formula
+  desc "A library and utilities for processing GIFs"
+  homepage "http://giflib.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.1.6/giflib-4.1.6.tar.bz2"
+  sha256 "e1c1ced9c5bc8f93ef0faf0a8c7717abf784d10a7b270d2285e8e1f3b93f2bed"
+
+  bottle do
```
DomT4: Kill this block. We'll add a new one on pull.

giflib41.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  keg_only "Conflicts with giflib in main repository."
+
+  def install
+    system "./configure", "--prefix=#{prefix}", "--disable-dependency-tracking"
+    system "make", "install"
+  end
+
+  test do
```
DomT4: Can we actually add a test here? Thanks. If it helps we ship a test `gif` which you can use, example [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/giflossy.rb#L38).

karmakaze: Ultimately, I think it's better to have the using code to update to 4.2 or 5 in my specific case and in general since it will require changes either way.

PR #889, opened 2015-07-09
antlr3: add formula for antlr3.4

bagage:
Antlr4 is the current main, but some people may need Antlr3, which has support for a C API.


antlr3.rb
```diff
@@ -0,0 +1,18 @@
+class Antlr3 < Formula
+  # Antlr4 is the current main, but some people may need Antlr3,
+  # which has support for a C API.
+  url "http://www.antlr3.org/download/antlr-3.4-complete.jar"
+  version '3.4'
```
DomT4: Double quotes everywhere.

antlr3.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage 'http://www.antlr.org/'
+  sha1 '5cab59d859caa6598e28131d30dd2e89806db57f'
```
DomT4: `sha256` here please.

antlr3.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def install
+    prefix.install "antlr-3.4-complete.jar"
```
DomT4: Would prefer to put this in `libexec` and fix the other `prefix` calls below.

antlr3.rb
```diff
@@ -0,0 +1,18 @@
...
+    (share+"java").install_symlink "#{prefix}/antlr-3.4-complete.jar" => "antlr3.jar"
+    (bin+"antlr-3.4").write <<-EOS.undent
+    #!/bin/sh
```
DomT4: Indent this two spaces so it's obvious it's not part of the main formula.

bagage:
Done - I'm not totally sure for `libexec` usage; is it ok now?

DomT4:
Changes look good, Thanks. Our test-bot flagged a few remaining issues:

```
==> audit problems
homebrew/versions/antlr3:
 * `homepage`(line 6) should be put before `url`(line 4)
 * Stable: version 3.4 is redundant with version scanned from URL
```

bagage:
Updated - I did not do a test: section though since I do not know what to test.

DomT4:
Presuming it works, you could steal the test from [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/antlr.rb#L16-L34).

DomT4:
Ping?

bagage:
My bad I totally forgot this. I tried to c/c the test but antlr 3.4 does not generate anything using the method of antlr4. I did not investigate why though - does this really matter or can we keep it without test?

DomT4:
Yeah, I can see why now I've looked at it. One of the tools used in that test looks like a new introduction to Antlr4.

There are some [examples](https://stackoverflow.com/questions/1931307/antlr-is-there-a-simple-example) at that link, It'd be nice to try and adapt them for this PR. It's troublesome to accept things without tests because we don't have a way of knowing the formula actually works at runtime without them.

bagage:
Finally done!

dunn:
Just one little thing left:
```
==> audit problems
homebrew/versions/antlr3:
Length is calculated as homebrew/versions/antlr3 + desc. (currently 124)
```

bagage:
Done again? This time everything is green! :)

antlr3.rb
```diff
@@ -0,0 +1,53 @@
+class Antlr3 < Formula
+  desc "Antlr3 has a C API which has been dropped in Antlr4."
+  homepage "http://www.antlr.org/"
+  url "http://www.antlr3.org/download/antlr-3.4-complete.jar"
+  sha256 "9d3e866b610460664522520f73b81777b5626fb0a282a5952b9800b751550bf7"
+
+  def install
+    libexec.install "antlr-3.4-complete.jar"
+    (share+"java").install_symlink "#{libexec}/antlr-3.4-complete.jar" => "antlr3.jar"
+    (bin+"antlr-3.4").write <<-EOS.undent
```
dunn: Why not just `(bin/"antlr3").write`, and skip line 14 below? Is there a reason to have the executable available under two different names?

bagage: indeed, fixed.

bagage:
Any news?

antlr3.rb
dunn: Do we want `bin/antlr3` or `bin/antlr-3.4`?

antlr3.rb
```diff
@@ -0,0 +1,52 @@
...
+      #!/bin/sh
+      java -jar #{libexec}/antlr-3.4-complete.jar "$@"
+      EOS
```
DomT4: `EOS` should line up with the start of `(bin+"antlr-3.4")`.write`

antlr3.rb
bagage: I think `bin/antlr-3.4` is better since `antlr-3.2` exists too.

antlr3.rb
bagage: aligned.

antlr3.rb
dunn: We don't have antlr 3.2 in this tap, though.  We do have antlr 2, and it's installed as `bin/antlr2`.

dunn: If you think we should have 3.2, then maybe `antlr34`?

bagage: Well if antlr2 is named as it, let's name it `antlr3` (`antlr3.2` is broken anyway, so nobody should use it!). Fixed.

bagage:
Bump?

DomT4:
@BrewTestBot test this please

DomT4:
@bagage Apologies, busy few days. If all is green on the test-bot this time as well I'll merge it shortly after.

bagage:
Sure, no prob. Thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/857638c6fbb7ab9b2d322d2dee4308f101647ce6. Thank you for your contribution to Versions, we appreciate it! :smiley_cat:

PR #891, opened 2015-07-09
add protobuf v2.6.0

edvorg:


protobuf260.rb
```diff
@@ -0,0 +1,144 @@
+class Protobuf < Formula
```
DomT4: You need to fix the `class` name to match the formula name.

DomT4:
You'll need to add a `conflicts_with "protobuf", :because => "Differing versions of same formula"` line here. Should be one blank line beneath the Python dependency.

edvorg:
fixed

DomT4:
Few more nits:

```
==> brew style protobuf260
== /usr/local/Library/Taps/homebrew/homebrew-versions/protobuf260.rb ==
W: 67: 31: Use parentheses in the method call to avoid confusion about precedence.
C:105:  7: Use 2 (not -9) spaces for indentation.
C:115:  5: Align else with if.
W:125:  5: end at 125, 4 is not aligned with if at 104, 15

1 file inspected, 4 offenses detected
```

protobuf260.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  def caveats; <<-EOS.undent
+    Editor support and examples have been installed to:
+      #{doc}
+    EOS
+  end
+
+  test do
+    testdata = if devel?
```
DomT4: You can remove all the `devel` calls. There won't be a `devel` in this case.

protobuf260.rb
```diff
@@ -0,0 +1,146 @@
...
+
+  def install
+    # Don't build in debug mode. See:
+    # https://github.com/Homebrew/homebrew/issues/9279
+    # http://code.google.com/p/protobuf/source/browse/trunk/configure.ac#61
+    ENV.prepend "CXXFLAGS", "-DNDEBUG"
+    ENV.universal_binary if build.universal?
+    ENV.cxx11 if build.cxx11?
+
+    system "./autogen.sh" if build.devel?
```
DomT4: And here.

edvorg:
fixed

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/e202abf467095cc26511e54f7be831f58ccdfe92. Thank you for your contribution to Versions, we appreciate it! :smiley_cat:.

Just a nit, but new formulae should have their commit message styled `protobuf260 2.6.0 (new formula)` or similar, Thanks!

PR #892, opened 2015-07-09
Created formula for installing drush version 6.5

jhuebsch:
Drush 7.0 has been released time to take 6.5 to the versions tap.


jhuebsch:
forgot to make keg only.

PR #916, opened 2015-07-17
influxdb 0.8.8

joel-airspring:
Influxdb 0.9 is the latest and greatest but has api changes, and not all tooling in the ecosystem supports it.

Could we please add a 0.8.8 version?

Thanks!

### DIFF
--- a/influxdb088.rb
+++ b/influxdb088.rb
@@ -1,0 +1,83 @@
+class Influxdb088 < Formula
+  desc "Time series, events, and metrics database"
+  homepage "http://influxdb.com"
+  url "https://s3.amazonaws.com/get.influxdb.org/influxdb-0.8.8.src.tar.gz"
+  sha256 "97fb5a4ffda1b333187ebd6449466d38d864686a3bd50a6c7bfb3deeae06cfae"
+
+  depends_on "leveldb"
+  depends_on "rocksdb"
+  depends_on "autoconf" => :build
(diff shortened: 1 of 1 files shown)


joel-airspring:
I'm going to tweak the name.

influxdb088.rb
```diff
@@ -0,0 +1,83 @@
...
+    inreplace "configure" do |s|
+      s.gsub! "echo -n", "$as_echo_n"
+    end
+
+    system "./configure", "--with-flex=#{flex}", "--with-bison=#{bison}", "--with-rocksdb"
+    system "make", "parser", "protobuf"
+    system "go", "build", "-tags", "leveldb rocksdb", "-o", "influxdb", "github.com/influxdb/influxdb/daemon"
+
+    inreplace "config.sample.toml" do |s|
+      s.gsub! "/tmp/influxdb/development/db", "#{var}/influxdb/data"
```
DomT4: If possible we should version the `var` path, so `#{var}/influxdb088/data` and similar.

We've been bitten fairly nastily by people switching around formula versions and then discovering one made a change to a data path that the other can't understand and now everything is on :fire: and people panic :wink:.

DomT4: Make sure it plays nicely with the `WorkingDirectory` line in the `plist` though. If it doesn't, I guess we'll just have to cope with having an un-versioned `var` path.

PR #920, opened 2015-07-20
Update README.md

abardagjy:
Updated the tap path in README.md from homebrew/versions to homebrew/homebrew-versions


DomT4:
Afraid `homebrew/versions` is actually correct here. Whilst Homebrew requires Github taps to start with `homebrew-` as long as that is present we don't need users to type it in the CLI. `brew tap homebrew/versions` and similar will work fine.

abardagjy:
Weird, bare `homebrew/versions` did not work for me on a new installation, and the verbose `homebrew/homebrew-versions` did. But then after untapping I cannot re-create the issue.

DomT4:
That is a little odd. Let us know if you hit the issue again!

PR #921, opened 2015-07-22
Remove root_url if bintray

chriskolenko:
Based on the this conversation: https://github.com/Homebrew/homebrew/issues/42008

I've removed any bintray root_urls

llvm33.rb && llvm34.rb still have root_urls pointing to downloads.sf.net (Should these be removed also?)


xu-cheng:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #922, opened 2015-07-22
Adding Hadoop 1.0.3

dgranahan:


DomT4:
```
==> audit problems
homebrew/versions/hadoop103:
Length is calculated as homebrew/versions/hadoop103 + desc. (currently 84)
```

DomT4:
Presuming dead. Feel free to resubmit with the issues addressed if desired, Thanks!

PR #927, opened 2015-07-26
fix broken brew relative path symlink

### LINKED ISSUE #926: bash-completion2: broken brew symlink
Current "brew" symlink created in post_install is linked to the wrong location because "/usr/local/share/bash-completion" is also a symlink to a directory in the different depth. To understand this issue, cd into /usr/local/share/bash-completion/completions, then run "ls ../../.." and "cd ../../..". You will see the different result. "ls ../../.." will go three level above the realpath, and show you the content in /usr/local/Cellar/bash-completion2/2.1_2/. But result of "cd ../../.." is three level above PWD, which is /usr/local.

carlosliu:
Fix #926


DomT4:
Why is this a problem here and not with `bash_competion` in the [core](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/bash-completion.rb)?

carlosliu:
The difference is that the change of completion files directory.

In the [core](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/bash-completion.rb), it is '/usr/local/etc/bash_completion.d', which is a realpath by itself.

In this version, the directory is '/usr/local/share/bash-completion2/completions' which is not a realpath. The '/usr/local/bash-completion2' part is a symlink pointed into Cellar/...

I've explained the difference between two situations in #926 . The relative path in a  symlink is only relative to the realpath destination.

DomT4:
@BrewTestBot test this please

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/71845e3280e37bb999ae6672e032d06fda49019e. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #932, opened 2015-07-29
Adding gradle24 formula.

gjroylance:
Gradle 2.4 distribution  formula added.

### DIFF
--- a/gradle24.rb
+++ b/gradle24.rb
@@ -1,0 +1,12 @@
+class Gradle24 < Formula
+  desc "Build system"
+  homepage "https://www.gradle.org/"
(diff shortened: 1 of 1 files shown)


gradle24.rb
```diff
@@ -0,0 +1,12 @@
+class Gradle24 < Formula
+  desc "Build system"
```
DomT4: This is a little general. Was there a problem with `Build system based on the Groovy language`?

DomT4:
Needs a `conflicts_with` line like [this](https://github.com/Homebrew/homebrew-versions/commit/c70582a2055ea6649cc1974076f57001f8c471a3).

PR #936, opened 2015-07-30
libmongoclient-legacy 1.0.4

billybones256:
Update to latest official driver.


dunn:
Is it possible to add a test as well? Just something to quickly check basic functionality.

billybones256:
I think no, because user need to have installed mongodb on his computer to pass tests.

dunn:
Something like this should work.  It doesn't do much, but it's better than nothing:
```ruby
  test do
    (testpath/"test.cpp").write <<-EOS.undent
      #include <mongo/client/dbclient.h>

      int main() {
          mongo::DBClientConnection c;
          mongo::client::initialize();
          return 0;
      }
    EOS
    system ENV.cxx, "-L#{lib}", "-lmongoclient",
           "-L#{Formula["boost"].opt_lib}", "-lboost_system",
           testpath/"test.cpp", "-o", testpath/"test"
    system "./test"
  end
```

billybones256:
Ok, I added you code to test section.

dunn:
Merged in bbf8b828f7f6e22993a7deaef0afd26e0993b2bc, thanks!

PR #937, opened 2015-08-07
Mapnik v2.2.0

jlnr:
### What

The last revision of the Homebrew Mapnik formula before v3.0.0, taken from here:

https://github.com/Homebrew/homebrew/blob/dd05e1c07f1f081cd758e96a47f1d546cc61d262/Library/Formula/mapnik.rb

I've removed the `revision`, `head` and `bottle`, added the `description` from the HEAD Mapnik formula, and added a (very, very) minimal test block.

### Why

- Mapnik 3 does not include the Python bindings anymore, a fact that isn't even mentioned yet in Mapnik's documentation. The new Python bindings cannot simple be installed via `pip`, quote: "Eventually we hope that many people will simply be able to pip install mapnik in order to get prebuilt binaries, this currently does not work though". This bottle still includes the Mapnik 2.x bindings.
- There are no pre-compiled downloads for Mapnik 3.x on the [official download page](http://mapnik.org/pages/downloads.html), so I have to use Mapnik 2.x for parity with the Windows developers on our team.
- Mapnik's documentation still claims that v2.2.0 is the latest stable version, and it just seems that the Mapnik ecosystem has not fully moved to v3.x yet.

### Why not Mapnik 2.3?

Because I am really new to Mapnik - I don't know much about all the moving parts 😅


mapnik2.rb
```diff
@@ -0,0 +1,104 @@
+class Mapnik2 < Formula
+  desc "Toolkit for developing mapping applications"
+  homepage "http://www.mapnik.org/"
+  url "https://s3.amazonaws.com/mapnik/dist/v2.2.0/mapnik-v2.2.0.tar.bz2"
+  sha256 "9b30de4e58adc6d5aa8478779d0a47fdabe6bf8b166b67a383b35f5aa5d6c1b0"
+
+  stable do
```
DomT4: Since there will never be anything but `stable` in this formula you can remove the block and move everything into the main line.

```ruby
homepage ""
url ""
sha256 ""

# Patch to fix example
patch :DATA

#  patch for something else
patch do
  url ""
  sha256 ""
end
```

jlnr:
Thanks @DomT4! I've force-pushed a fix for the whitespace that the build bot has found and removed the `stable` block.

DomT4:
I heavily suspect this is going to break over time. Boost has this habit of breaking everything from time to time, so we may end up having to chain this to an older boost in this tap to ensure it keeps working, but it's passing at the moment so will merge as-is.

Merged in https://github.com/Homebrew/homebrew-versions/commit/a4f62e9e2c1dbc4e2e7ee2211179c1dc46138e70. Thank you for your contribution to Versions, we appreciate it! :smiley_cat:

PR #938, opened 2015-08-11
Add jruby files for the current and previous two minor versions.

booc0mtaco:
Adding the latest release of jruby (9.0.0.0), plus the two previous patch versions for the 1.7.x and 1.6.x versions (21 and 8 respectively). 

Referenced the sha1 values listed on the jruby.org site.

Please let me know if there are any changes that should be made in case this isn't formatted exactly right.


jruby168.rb
```diff
@@ -0,0 +1,30 @@
+class Jruby < Formula
```
dunn: The classname needs to match the filename; `Jruby168` and so on.

jruby168.rb
```diff
@@ -0,0 +1,30 @@
...
+  desc "Ruby implementation in pure Java"
+  homepage "http://www.jruby.org"
+  url "https://s3.amazonaws.com/jruby.org/downloads/1.6.8/jruby-bin-1.6.8.tar.gz"
+  sha1 "b1d191723b34255e5833946c6591bc5e89379d8c"
```
DomT4: These should all be `sha256`.

jruby168.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  depends_on :java => "1.7+"
```
DomT4: One empty line underneath this you need to `conflicts_with` the core & each added `jruby`, like this:

```ruby
depends_on :java => "1.7+"

conflicts_with "jruby", :because => "Differing version of the same formula"
conflicts_with "jruby1721", :because => "Differing version of the same formula"
```

And so on.

DomT4: (The change above needs adding to all the proposed formulae here, not just one, just to be clear)

jruby168.rb
booc0mtaco: OK, I'll make the changes first thing tomorrow. I can redo this PR so it's all in one commit, if you prefer. Otherwise, I'll make another commit with my fixes in.

dunn: Feel free to amend your original commit and force push to this branch, too.

booc0mtaco:
Hey there, 

looks like tests have passed. Anything else I should do here? Let me know !

DomT4:
Ideally it would have been done in three separate commits, one for each formula, but unless you want to do that in the next couple hours I'll pull this as-is later this evening.

booc0mtaco:
Ah, I see. Should I have done three commits in one PR, or one commit per PR? Just want to be clear for next time (I probably won't get a chance to touch this until a bit later)

DomT4:
All in the same PR is fine, but one commit per formula added with the commit message something like:

```
jruby168 1.6.8 (new formula)
```

booc0mtaco:
Ok, I've got a few minutes, so should have this up shortly.

DomT4:
Merged! Thank you for your contribution to Versions, and speedy response to the various feedback on this. Really appreciate that.

PR #939, opened 2015-08-13
Docker 1.7.1

RevCBH:
I disabled the bottles section because I don't know what that does and it seems to work without it. Let me know what to do if I should re-enable them.

### DIFF
--- a/docker171.rb
+++ b/docker171.rb
@@ -1,4 +1,4 @@
-class Docker162 < Formula
+class Docker171 < Formula
   desc "The Docker framework for containers"
   homepage "https://www.docker.com/"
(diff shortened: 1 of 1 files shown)


mistydemeo:
You can delete that section entirely. It refers to Homebrew's binary packages.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/2b8fa8d6c3977aecc45e527dc7e99f781f361f6f. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but the commit message for new formula should be styled `docker171 1.7.1 (new formula)`, Thanks!

PR #944, opened 2015-08-19
elasticsearch 1.6.2 (new formula)

patricknmahoney:
Elasticsearch 1.6

I put it in as a PR since it will function as-is currently, but I'm also amenable to adding the checksums for other os x versions (guidance would be appreciated). In the meantime, this would be useful to start.


patricknmahoney:
nnnnooope

PR #945, opened 2015-08-19
elasticsearch 1.6.2 (new formula)

patricknmahoney:
Elasticsearch 1.6.2 -- note the branch on my fork is called elasticsearch16-2; I can change that if that convention matters. I've tested this one with 1.6.2 and `brew install homebrew/versions/elasticsearch16`.

One question -- is this being done redundantly wherein the cellar install ends up being `/usr/local/Cellar/elasticsearch16/1.6.2` ; this seems wrong but also is in line with the output from `brew install homebrew/versions/elasticsearch14` (which installs, similarly, to `/usr/local/Cellar/elasticsearch14/1.4.5`); neither of which can guarantee the user's symlinks are safely in place.

So 2 additional notes:

1.  I put it in as a PR since it will function as-is currently, but I'm also amenable to adding the checksums for other os x versions (guidance would be appreciated). In the meantime, my guess is this would be useful to start.

2.  If anyone has thoughts on whether it makes more sense to utilize the previously-included-as-canonical version of elasticsearch (1.6.0 was in the main homebrew formulae), as opposed to this one (the stable release, 1.6.2), I'm eager to hear your thoughts!

```bash
brew install homebrew/versions/elasticsearch16 
==> Installing elasticsearch16 from homebrew/homebrew-versions
==> Downloading https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.6.2.tar.gz
######################################################################## 100.0%
==> Caveats
Data:    /usr/local/var/elasticsearch/elasticsearch_patricknavarromahoney/
Logs:    /usr/local/var/log/elasticsearch/elasticsearch_patricknavarromahoney.log
Plugins: /usr/local/var/lib/elasticsearch/plugins/
Config:  /usr/local/etc/elasticsearch/

To reload homebrew/versions/elasticsearch16 after an upgrade:
  launchctl unload ~/Library/LaunchAgents/homebrew.mxcl.elasticsearch16.plist
  launchctl load ~/Library/LaunchAgents/homebrew.mxcl.elasticsearch16.plist
Or, if you don't want/need launchctl, you can just run:
  elasticsearch --config=/usr/local/opt/elasticsearch16/config/elasticsearch.yml
==> Summary
🍺  /usr/local/Cellar/elasticsearch16/1.6.2: 34 files, 30M, built in 11 seconds
```


DomT4:
>  note the branch on my fork is called elasticsearch16-2; I can change that if that convention matters

I frequently name my branches after songs; feel free to call your branches whatever you like, within reason :smile:.

> One question -- is this being done redundantly wherein the cellar install ends up being

Everything gets installed into `HOMEBREW_CELLAR/name/version/*`. This is standard for the whole of Homebrew.

> neither of which can guarantee the user's symlinks are safely in place.

I don't quite get the question here?

> If anyone has thoughts on whether it makes more sense to utilize the previously-included-as-canonical version of elasticsearch

Given we've called this `elasticsearch16` rather than `elasticsearch160` it should always be the latest release on that branch, which is in line with the expectation of the other `elasticsearch` formula in Versions here. If `1.6.2` is the latest `1.6.x` that's great.

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
+class Elasticsearch16 < Formula
+  desc "Distributed real-time search & analytics engine for the cloud"
+  homepage "https://www.elastic.co/products/elasticsearch"
+  url "https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.6.2.tar.gz"
+  sha256 "b7ef3aae0a263c2312bd1a25b191c3c108c92d5413c3527d776587e582c518d0"
+
+  bottle do
```
DomT4: You can remove this whole block. We don't actually add bottles for `elasticsearch` as it doesn't benefit the user in terms of speed or ease of use at all, as the process of both is download files => dump files in prefix. There's no compile or build dependencies being saved.

DomT4: Side note: Even if we did generate bottles, our CI adds them automatically and they don't need adding manually.

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+    cellar :any
+    sha256 "b7ef3aae0a263c2312bd1a25b191c3c108c92d5413c3527d776587e582c518d0" => :yosemite
+  end
+
+  depends_on :java => "1.7+"
+
+  def cluster_name
+    "elasticsearch_#{ENV["USER"]}"
```
DomT4: I threw together a rule a while back to handle conflicting versions of data directories, after making changes broke a bunch of stuff for users. Every new formula that writes data to `var` (which consequently sticks around between removals, unlinks, etc) and can conflict with other formulae should have their data directories versioned according to the formula name.

In this case, ideally `elasticsearch16_#{ENV["USER"]}`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+    # Remove unnecessary files
+    rm_f Dir["#{lib}/sigar/*"]
+
+    # Set up Elasticsearch for local development:
+    inreplace "#{prefix}/config/elasticsearch.yml" do |s|
+      # 1. Give the cluster a unique name
+      s.gsub!(/#\s*cluster\.name\: elasticsearch/, "cluster.name: #{cluster_name}")
+
+      # 2. Configure paths
+      s.sub!(%r{#\s*path\.data: /path/to.+$}, "path.data: #{var}/elasticsearch/")
```
DomT4: Same gig here. `#{var}/elasticsearch16`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+      s.sub!(%r{#\s*path\.logs: /path/to.+$}, "path.logs: #{var}/log/elasticsearch/")
```
DomT4: `#{var}/log/elasticsearch16/`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+      s.sub!(%r{#\s*path\.plugins: /path/to.+$}, "path.plugins: #{var}/lib/elasticsearch/plugins")
```
DomT4: `#{var}/lib/elasticsearch16/plugins`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    # Move config files into etc
+    (etc/"elasticsearch").install Dir[prefix/"config/*"]
+    (prefix/"config").rmtree
+  end
+
+  def post_install
+    # Make sure runtime directories exist
+    (var/"elasticsearch/#{cluster_name}").mkpath
+    (var/"log/elasticsearch").mkpath
```
DomT4: `(var/"log/elasticsearch16").mkpath`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+    (var/"lib/elasticsearch/plugins").mkpath
```
DomT4: `(var/"lib/elasticsearch16/plugins").mkpath`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    inreplace "#{bin}/plugin" do |s|
+      # Add the proper ES_CLASSPATH configuration
+      s.sub!(/SCRIPT="\$0"/, %(SCRIPT="$0"\nES_CLASSPATH=#{libexec}))
+      # Replace paths to use libexec instead of lib
+      s.gsub!(%r{\$ES_HOME/lib/}, "$ES_CLASSPATH/")
+    end
+
+    # Move config files into etc
+    (etc/"elasticsearch").install Dir[prefix/"config/*"]
```
DomT4: `(etc/"elasticsearch16")`

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+    (prefix/"config").rmtree
+  end
+
+  def post_install
+    # Make sure runtime directories exist
+    (var/"elasticsearch/#{cluster_name}").mkpath
+    (var/"log/elasticsearch").mkpath
+    (var/"lib/elasticsearch/plugins").mkpath
+    ln_s etc/"elasticsearch", prefix/"config"
```
DomT4: Can you check if `prefix.install_symlink etc/"elasticsearch16" => "config"` works?

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  def caveats; <<-EOS.undent
+    Data:    #{var}/elasticsearch/#{cluster_name}/
```
DomT4: And these should be `elasticsearch16` where stated above as well.

elasticsearch16.rb
```diff
@@ -0,0 +1,119 @@
...
+      <plist version="1.0">
+        <dict>
+          <key>KeepAlive</key>
+          <true/>
+          <key>Label</key>
+          <string>#{plist_name}</string>
+          <key>ProgramArguments</key>
+          <array>
+            <string>#{HOMEBREW_PREFIX}/bin/elasticsearch</string>
+            <string>--config=#{etc}/elasticsearch/elasticsearch.yml</string>
```
DomT4: `#{etc}/elasticsearch16/elasticsearch.yml`

patricknmahoney:
Very cool! I just saw this, and I appreciate you explaining. Good point re the convention/rule you outline makes it easier for folks to stay consistent with that, and not have that mess of data folders and config symlinked-to-the-lib (->sneaky and hard-to-diagnose bugs e.g. things that would be inconsistently linked/accessed, potentially running configs for wrong versions).

I'll refactor this PR later today, based on the notes you've left here! Thanks

PR #948, opened 2015-08-25
libmongoclient-legacy 1.0.5

billybones256:
Update official legacy driver to version 1.0.5.


DomT4:
Could you possibly squash your commits back down to 1 here please, Thanks!

PR #949, opened 2015-08-25
libxc2.1 (new formula)

tmarkovich:
I would like to build and distribute octopus via homebrew, but octopus requires building libxc-2.1.0 instead of libxc-latest, so I created this specific version formula. I'll commit the octopus formula in just a bit. This is in an effort to build and distribute an open source electronic structure package.

Octopus can be found at tddft.org


mistydemeo:
@BrewTestBot test this please

mistydemeo:
@BrewTestBot test this please

PR #956, opened 2015-09-02
Add Cassandra 2.1 Formula

cscetbon:


cscetbon:
sorry I can't let your bot access my organization

tdsmith:
FWIW the bot asks for organization and team membership to decide if you're a member of a Homebrew team.

The error is
```
== /usr/local/Library/Taps/homebrew/homebrew-versions/cassandra21.rb ==
C: 86: 15: Use %r around regular expression.

1 file inspected, 1 offense detected
Error: 2 problems in 1 formula

==> audit problems
homebrew/versions/cassandra21:
Length is calculated as homebrew/versions/cassandra21 + desc. (currently 81)
```

tdsmith:
I'm not sure if that feedback is actionable.

cscetbon:
@tdsmith should be better now

cassandra21.rb
```diff
@@ -0,0 +1,117 @@
+class Cassandra21 < Formula
+  desc "Eventually consistent distributed key-value store"
+  homepage "https://cassandra.apache.org"
+  url "https://www.apache.org/dyn/closer.cgi?path=/cassandra/2.1.8/apache-cassandra-2.1.8-bin.tar.gz"
+  mirror "https://archive.apache.org/dist/cassandra/2.1.8/apache-cassandra-2.1.8-bin.tar.gz"
+  sha256 "3a0cc64efd529ffdc1600f6b3ad1946af85cc01544e2b469499aa81b10b722f5"
+
+  bottle do
```
DomT4: Please don't add this block. We'll add one on pull automatically.

cassandra21.rb
```diff
@@ -0,0 +1,117 @@
...
+    sha256 "7857c16d8aab7b736ab677d1016ef8513dedb64097214ad3a50a6c550cb7d6e0"
+  end
+
+  resource "cassandra-driver" do
+    url "https://pypi.python.org/packages/source/c/cassandra-driver/cassandra-driver-2.5.1.tar.gz"
+    sha256 "98511b51f3565bf5836079ed31b7be681d5a17b5a98d1214aa64eb06e07bc18d"
+  end
+
+  def install
+    (var+"lib/cassandra").mkpath
```
DomT4: Please version these directories. Even with a `conflicts_with` we can't unlink the data directories and we've had plenty of issues where the data has used the same paths and caused incompatibility chaos.

cassandra21.rb
```diff
@@ -0,0 +1,117 @@
...
+    inreplace "bin/cassandra.in.sh" do |s|
+      s.gsub! "CASSANDRA_HOME=\"`dirname \"$0\"`/..\"", "CASSANDRA_HOME=\"#{libexec}\""
+      # Store configs in etc, outside of keg
+      s.gsub! "CASSANDRA_CONF=\"$CASSANDRA_HOME/conf\"", "CASSANDRA_CONF=\"#{etc}/cassandra\""
+      # Jars installed to prefix, no longer in a lib folder
+      s.gsub! "\"$CASSANDRA_HOME\"/lib/*.jar", "\"$CASSANDRA_HOME\"/*.jar"
+      # The jammm Java agent is not in a lib/ subdir either:
+      s.gsub! "JAVA_AGENT=\"$JAVA_AGENT -javaagent:$CASSANDRA_HOME/lib/jamm-", "JAVA_AGENT=\"$JAVA_AGENT -javaagent:$CASSANDRA_HOME/jamm-"
+      # Storage path
+      s.gsub! "cassandra_storagedir\=\"$CASSANDRA_HOME/data\"", "cassandra_storagedir\=\"#{var}/lib/cassandra\""
```
DomT4: Same note here on the versioning.

cassandra21.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    rm Dir["bin/*.bat", "bin/*.ps1"]
+
+    # This breaks on `brew uninstall cassandra && brew install cassandra`
+    # https://github.com/Homebrew/homebrew/pull/38309
+    (etc+"cassandra").install Dir["conf/*"]
```
DomT4: And here, re versioning.

DomT4:
As hinted at in the commit comment, need a `conflicts_with "cassandra", :because => "Differing version of same formula"` line one empty line beneath the last dependency.

cassandra21.rb
cscetbon: @DomT4 :ok_hand:

cassandra21.rb
cscetbon: @DomT4 it's based on formula cassandra20. If it's a new requirement I'll fix it

cassandra21.rb
cscetbon: @DomT4 I guess I have to do it everywhere !

cscetbon: @DomT4 I'm thinking that if we use versioned directories, it will mean that users won't be able to update it ! I don't think it's a good idea. What do you think ? users could want to check if they can update from 2.0 to 2.1 for example.

cscetbon: @DomT4 I was speaking about the data directory not the other ones

cscetbon:
@DomT4 shouldn't it conflict with all other formulas version ?
```
cassandra-0.6.rb
cassandra10.rb
cassandra12.rb
cassandra20.rb
```

cassandra21.rb
DomT4: We didn't "fix" the existing formulae because that had the potential to cause silent breakage for existing users. Since that isn't a new problem for new formulae, we have started versioning data directories to stop cross-version-contamination.

cassandra21.rb
DomT4: I'm not sure quite what you mean, but they won't be able to update it?

cscetbon: @DomT4 yes they won't be able to update their data if we use different data directories. However we could only use different version for directories other than data. Ok for you ?

apjanke:
Ideally, yes, but we don't usually put all the versioned conflicts in. Just the one against the core formula is required. It's nice if the others are added, too.

MikeMcQuaid:
Closing this but will accept a new PR with all comments addressed. Thanks!

PR #966, opened 2015-09-14
gnupg21 2.1.8

DomT4:


chdiza:
According to the release notes for 2.1.8 there's no longer a `libksba` dep:

```
gpg: Does not anymore link to Libksba to reduce dependencies.
```

DomT4:
Nice spot @chdiza. If there's no linkage there I'll just remove the dependency after pull. Will double check the bottle hasn't opportunistically linked first.

DomT4:
Garg, there's still some linkage there:

```
gnupg21/2.1.8/bin/dirmngr
gnupg21/2.1.8/bin/gpgsm
gnupg21/2.1.8/bin/kbxutil
```

DomT4:
Huh. It didn't build:

```
configure:
***
*** You need libksba to build this program.
*** This library is for example available at
***   ftp://ftp.gnupg.org/gcrypt/libksba/
*** (at least version 1.2.0 using API 1 is required).
***
configure: error: 
***
*** Required libraries not found. Please consult the above messages
*** and install them before running configure again.
***
```

DomT4:
Closed via https://github.com/Homebrew/homebrew-versions/commit/caf2cde356075772523ae839c969ffd909ce87aa.

Will look at the `ksba` problem going forwards, but if anyone else wants to poke upstream and ask why configure fails without it, that'd be cool.

chdiza:
>  but if anyone else wants to poke upstream and ask why configure fails without it, that'd be cool.

Alas, it won't be me.  I don't even use 2.1.8, and I'm not gonna test it because I don't want to inadvertently mess up my `~/.gnupg`, what with the new keychain/box format stuff.

PR #967, opened 2015-09-16
Add Python 3.4.3

kudos:
Lifted from the main homebrew repo, but removing 3.5 references from skip clean and adding a conflicts check for other python3 being installed.


tdsmith:
Please use pyenv or this tap for historic Python versions: https://github.com/zoidbergwill/homebrew-python

Thanks!

PR #970, opened 2015-09-21
Added Hadoop 2.4.0.

darrin:
Simple script to install old/archived version of Hadoop from archive.apache.org. Hive 2.4.0 is the LATEST version available on Amazon EMR AMIs so we cannot be the only ones that still want access to this!

### DIFF
--- a/hadoop240.rb
+++ b/hadoop240.rb
@@ -1,0 +1,37 @@
+class Hadoop240 < Formula
+  desc "Framework for distributed processing of large data sets"
+  homepage 'https://hadoop.apache.org/'
+  url 'https://archive.apache.org/dist/hadoop/core/hadoop-2.4.0/hadoop-2.4.0.tar.gz'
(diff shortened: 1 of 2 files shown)


dunn:
Some audit failures for Hadoop:
```
== /usr/local/Library/Taps/homebrew/homebrew-versions/hadoop240.rb ==
C:  3: 12: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  4:  7: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  5:  8: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

1 file inspected, 3 offenses detected
Error: 3 problems in 1 formula

==> audit problems
homebrew/versions/hadoop240:
Length is calculated as homebrew/versions/hadoop240 + desc. (currently 84)
```
and Hive:
```
==> audit problems
homebrew/versions/hive0131:
```

DomT4:
Closing as dead. Please feel free to resubmit with all the things Alex highlighted fixed, Thanks.

PR #971, opened 2015-09-22
move iojs3 formula to versions

chrmoritz:
This PR moves the iojs forumula from homebrew core to versions as suggested in https://github.com/Homebrew/homebrew/pull/44147#issuecomment-141873655.
 
It also backports some recent improvements to the node forumula like adding support for statically linking `icu4c` and shipping with `npm` 2.14.4.

For easier review I created a gist with the [diff from the current iojs formula](https://gist.github.com/chrmoritz/47ae144f77d8f73ba04e/revisions#diff-52bef6ece3ab01bba7f35cd1f15e8879) in core and a [diff from the current node formula](https://gist.github.com/chrmoritz/c19cb4ccd517bb503d97/revisions#diff-52bef6ece3ab01bba7f35cd1f15e8879).

Note: The iojs formula should be removed from homebrew core when merging this.

### DIFF
--- a/iojs.rb
+++ b/iojs.rb
@@ -1,0 +1,135 @@
+class Iojs < Formula
+  desc "npm-compatible platform based on Node.js"
+  homepage "https://iojs.org/"
+  url "https://iojs.org/dist/v3.3.1/iojs-v3.3.1.tar.xz"
+  sha256 "c5b1a7117b75dca563f66c35ee65db6fc43e25a3756608ad2c4d60087314bd36"
+
+  option "with-debug", "Build with debugger hooks"
+  option "without-npm", "npm will not be installed"
+  option "without-completion", "npm bash completion will not be installed"
+  option "with-full-icu", "Build with full-icu (all locales) instead of small-icu (English only)"
+
+  depends_on "pkg-config" => :build
+  depends_on "openssl" => :optional
+  depends_on :python => :build if MacOS.version <= :snow_leopard
+
+  conflicts_with "node", :because => "Differing release line of the node formula"
+
+  fails_with :llvm do
+    build 2326
+  end
+
+  resource "npm" do
+    url "https://registry.npmjs.org/npm/-/npm-2.14.4.tgz"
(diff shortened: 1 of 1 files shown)


chrmoritz:
```
==> brew doctor
==> FAILED
Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:

    brew install numpy
```

Build error on mavericks looks unrelated.

DomT4:
Let's keep the formula name as `iojs`. That way I can "migrate" it directly from the core & existing `iojs` users will continue to get updates without having to hunt down the new formula & wondering why the existing one never updates any more. The CI failure is indeed unrelated.

iojs3.rb
```diff
@@ -0,0 +1,137 @@
...
+  homepage "https://iojs.org/"
+  url "https://iojs.org/dist/v3.3.1/iojs-v3.3.1.tar.xz"
+  sha256 "c5b1a7117b75dca563f66c35ee65db6fc43e25a3756608ad2c4d60087314bd36"
+
+  option "with-debug", "Build with debugger hooks"
+  option "without-npm", "npm will not be installed"
+  option "without-completion", "npm bash completion will not be installed"
+  option "with-full-icu", "Build with full-icu (all locales) instead of small-icu (English only)"
+
+  deprecated_option "with-icu4c" => "with-full-icu"
```
DomT4: You can kill the `deprecated_option`. Nothing is deprecated in a "new" formula, which this technically is :)

chrmoritz:
I've renamed it to `iojs` and removed the `deprecated_option`. Now it fails because of the name conflict with `iojs` in core, which shouldn't be a problem because it's going to be removed from there anyway.

chrmoritz:
@DomT4 Could you please take another look at this.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/624aeed3a334446472c3982ae977a0fc176e73bf. Thanks for getting this done @chrmoritz! :beer:

PR #974, opened 2015-09-27
saxon6: New Formula

felixbuenemann:
This adds a new formula for the old, but still popular version 6 of the SAXON XSLT processor. This version was the last before the split into the open source Saxon-B and the commercial Saxon-SA (now Saxon-HE/PE/EE/CE). It only support XSLT 1.0, but is featured eg. in the recently released book [XSLT Jumpstarter](https://pragprog.com/book/djkxsl/xslt-jumpstarter) and is has a much different/simpler commandline syntax than the newer versions so I think it's useful to have it in versions.

The formula installs a `saxon6` command, so this does not clash with the existing version of saxon in homebrew.


felixbuenemann:
Hmm, I adapted this from the original saxon formula, which is also lacking a test. Should be rather simple to add one.

felixbuenemann:
OK, test added,

felixbuenemann:
Updated to `.test(&block)` syntax.

felixbuenemann:
Swapped version and sha256 lines, to make `brew audit --strict` happy.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/6031287ca77af761e66bb5d5d3c33d873c3802da. Thank you for your contribution to Versions @felixbuenemann! :smiley_cat:

PR #989, opened 2015-10-08
android-ndk-r10d (new formula)

dalinaum:
It is based on https://github.com/Homebrew/homebrew/commit/074012a831a8d8d7bfd69d3fefb5a089e97add45 and https://github.com/Homebrew/homebrew/commit/8dd3478131b232553f4c0cab197f8443a0a559a6


cmelchior:
AFAIK we need to upgrade the NDK version used by core as well before we can merge this.

dalinaum:
`These object files were linked against the deprecated system OpenSSL.`

How can I resolve this problem?

dunn:
If these are pre-compiled executables, probably nothing.  We might have to ignore the audit in this case.

dalinaum:
Actually I added the following simple test.

```
test do
  shell_output("#{bin}/ndk-build --help > test.txt")
  assert File.read("test.txt").start_with?("Usage")
end
```

How about adding this @DomT4 ?

DomT4:
@dalinaum Does:

```ruby
test do
  assert_match /Usage/, shell_output("#{bin}/ndk-build --help)
end
```
Not work?

dalinaum:
@DomT4 Thanks. It works!

PR #994, opened 2015-10-19
Add formula for physfs 1.x.

wizonesolutions:
Needed to build ManaPlus on OS X.

### DIFF
--- a/physfs1.rb
+++ b/physfs1.rb
@@ -1,0 +1,33 @@
+class Physfs1 < Formula
+  desc "Library to provide abstract access to various archives"
+  homepage "https://icculus.org/physfs/"
+  url "https://icculus.org/physfs/downloads/physfs-1.1.1.tar.gz"
+  sha256 "994b65066657d49002509c9d84a9b0d2f5c0c1dab0457e9fe21ffa40f2205f63"
+
+  conflicts_with "physfs", :because => "Differing versions of the same formula"
+
+  head "https://hg.icculus.org/icculus/physfs/", :using => :hg
+
+  depends_on "cmake" => :build
+  depends_on "readline" => :test
+
+  def install
+    mkdir "macbuild" do
+      args = std_cmake_args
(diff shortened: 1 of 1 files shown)


wizonesolutions:
Hmm; it installed locally. How can I test/fix these failures?

DomT4:
```
==> audit problems
homebrew/versions/physfs1:
Length is calculated as homebrew/versions/physfs1 + desc. (currently 81)
 * Dependency readline does not define option :test
 * Dependency readline does not define option :test
```

physfs1.rb
```diff
@@ -0,0 +1,33 @@
...
+  homepage "https://icculus.org/physfs/"
+  url "https://icculus.org/physfs/downloads/physfs-1.1.1.tar.gz"
+  sha256 "994b65066657d49002509c9d84a9b0d2f5c0c1dab0457e9fe21ffa40f2205f63"
+
+  conflicts_with "physfs", :because => "Differing versions of the same formula"
+
+  head "https://hg.icculus.org/icculus/physfs/", :using => :hg
+
+  depends_on "cmake" => :build
+  depends_on "readline" => :test
```
DomT4: You can't have `test` only dependencies, essentially.

physfs1.rb
```diff
@@ -0,0 +1,33 @@
+class Physfs1 < Formula
+  desc "Library to provide abstract access to various archives"
+  homepage "https://icculus.org/physfs/"
+  url "https://icculus.org/physfs/downloads/physfs-1.1.1.tar.gz"
+  sha256 "994b65066657d49002509c9d84a9b0d2f5c0c1dab0457e9fe21ffa40f2205f63"
+
+  conflicts_with "physfs", :because => "Differing versions of the same formula"
+
+  head "https://hg.icculus.org/icculus/physfs/", :using => :hg
```
DomT4: Versions formulae don't have `head`s unless it points specifically to the branch of the version it is targeting.

PR #998, opened 2015-10-26
Add formula for ElasticSearch 1.5.2.

scraton:
AWS recently released [ElasticSearch Service](https://aws.amazon.com/elasticsearch-service/). The version they currently run is 1.5.2, so having this installable via homebrew would be nice. :)

### DIFF
--- a/elasticsearch15.rb
+++ b/elasticsearch15.rb
@@ -1,5 +1,5 @@
 class Elasticsearch15 < Formula
-  desc "Distributed real-time search & analytics engine for the cloud"
+  desc "You know, for search..."
   homepage "https://www.elastic.co/products/elasticsearch"
   url "https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.5.2.tar.gz"
   sha256 "efae7897367ac3fa8057d02fad31c72e215b6edef599b63e373d3ce0c1049a14"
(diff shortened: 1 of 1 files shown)


scraton:
Sorry for all the extra commits. Long time leacher, first time contributor. :sweat_smile:

elasticsearch15.rb
```diff
@@ -0,0 +1,133 @@
+class Elasticsearch15 < Formula
+  desc "You know, for search..."
```
DomT4: Let's stick to the Homebrew description here:

```ruby
desc "Distributed real-time search & analytics engine for the cloud"
```

elasticsearch15.rb
```diff
@@ -0,0 +1,133 @@
...
+  homepage "https://www.elastic.co/products/elasticsearch"
+  url "https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.5.2.tar.gz"
+  sha256 "efae7897367ac3fa8057d02fad31c72e215b6edef599b63e373d3ce0c1049a14"
+
+  bottle do
```
DomT4: Don't need this block here.

elasticsearch15.rb
scraton: Even though it causes the build/validations to fail?

elasticsearch15.rb
scraton: Copy that. Will remove.

elasticsearch15.rb
DomT4: Shorten it a bit if necessary.

scraton:
Applied the requested changes. :)

petercm:
+1
Though I'd bet you'll need to squash those commits

mikemcquaid:
Out of interest: are there specific breaking changes that mean you want to be on the identical version to AWS? We're currently rethinking the way we do Homebrew versions to being +/-2 versions from core.

scraton:
@mikemcquaid I'm not aware of any specific incompatibilities between the1.5 and 1.7. But I felt it'd be wise to develop against what AWS has to avoid any unpleasant surprises.

If you need me to squash these commits and open a new PR, just let me know. :)

mikemcquaid:
Unfortunately we've remove all versions below 1.7 as they are now unmaintained by upstream: https://github.com/Homebrew/homebrew-versions/pull/1068. Sorry!

PR #999, opened 2015-10-28
NiFi 0.2.1

apiri:
Adding a formula for a version 0.2.1 of Apache NiFi


nifi021.rb
```diff
@@ -0,0 +1,25 @@
+class Nifi < Formula
```
DomT4: You need to fix the `class` name to match the formula name.

apiri: Updated and pushed the change.  Thanks!

DomT4:
Just an audit nit left:

```
==> audit problems
homebrew/versions/nifi021:
Length is calculated as homebrew/versions/nifi021 + desc. (currently 101)
```

apiri:
Patched up, apologies for the audit miss.  Haven't quite worked out the best way to work on taps in terms of formula creations and wasn't able to find any good docs on that area.  Any suggestions as to how this is best accomplished and/or if I am overlooking something obvious?

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/b9fb9eacb1790d7ba290dd480f784e528f5d2e7b. Thank you for your contribution to Homebrew @apiri; we appreciate it! :smiley_cat: 

If you have specific questions on tap things let me know - Happy to answer. The easiest way to check everything is to run `brew audit --strict --online /path/to/formula.rb` on any new submissions.

PR #1006, opened 2015-11-03
add formula for elixir 1.0

ntrepid8:
This adds the formula for `elixir10` with version `1.0.5` since homebrew is now using Elixir version `1.1.x` and some libraries still need the older version.

This is my first PR against `homebrew-versions` so please let me know if anything needs to be changed for a merge.  I ran `brew audit --scrict` and it seem to be ok.  Thanks!

### DIFF
--- a/elixir10.rb
+++ b/elixir10.rb
@@ -1,0 +1,24 @@
+class Elixir10 < Formula
+  desc "Functional/dynamic language built on Erlang VM"
+  homepage "http://elixir-lang.org/"
+  url "https://github.com/elixir-lang/elixir/archive/v1.0.5.tar.gz"
+  sha256 "5ce5c226b3d11d751b41ad79b915b86f13f8a1b89ef3e733321d3f46ff4d81b8"
+
+  depends_on "erlang"
+
+  keg_only "Conflicts with elixir in main repository"
+
+  def install
+    system "make"
+    bin.install Dir["bin/*"] - Dir["bin/*.{bat,ps1}"]
+
+    Dir.glob("lib/*/ebin") do |path|
+      app = File.basename(File.dirname(path))
+      (lib/app).install path
+    end
+  end
+
+  test do
+    system "#{bin}/elixir", "-v"
+  end
+end


ntrepid8:
Why does the homebrew bot need access to read all my data, both for my account and every team that I am a member of?  Am I reading that wrong?

Is it going to just keep failing the build unless I give it permission?

screenshot 2015-11-03 18 18 59

tdsmith:
It wants to know if you're a member of the Homebrew organization. This doesn't grant the bot access to repositories. (see [scopes](https://developer.github.com/v3/oauth/#scopes))

The failure is unrelated:
```
==> audit problems
homebrew/versions/elixir10:
Length is calculated as homebrew/versions/elixir10 + desc. (currently 88)
```

ntrepid8:
Hey, thanks for the quick reply.  Those permissions don't seem like they would be an issue, it just freaked me out at first.  I've got it approved now and I'll look at the error.

ntrepid8:
So far I can't reproduce the error:

```
$ brew audit elixir10 --strict --online
==> brew style elixir10

1 file inspected, no offenses detected
```

I did see this issue: https://github.com/Homebrew/linuxbrew/issues/451

I'm still trying to reproduce the error that Jenkins is seeing but by my manual count that string is only 60 characters long and is exactly the same as the formula for Elixir in homebrew.

elixir10.rb
```diff
@@ -0,0 +1,50 @@
+class Erlang17Requirement < Requirement
```
DomT4: @dunn Would replacing this giant block with `depends_on :emacs => "17"` work?

elixir10.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+end
+
+class Elixir10 < Formula
+  desc "Functional metaprogramming aware language built on Erlang VM"
+  homepage "http://elixir-lang.org/"
+  url "https://github.com/elixir-lang/elixir/archive/v1.0.5.tar.gz"
+  sha256 "5ce5c226b3d11d751b41ad79b915b86f13f8a1b89ef3e733321d3f46ff4d81b8"
+
+  bottle do
```
DomT4: Can kill this block. We'll add bottles on pull.

ntrepid8:
I think I need to make one more update because this might need to be `keg_only` to avoid a conflict.  Checking...

ntrepid8:
I think it's good to go now.  Thanks for all the feedback on my first PR to this repo.  If there is anything else that needs to be updated please let me know.

DomT4:
@BrewTestBot test this please

ntrepid8:
Any more thoughts on this PR?

mikemcquaid:
@ntrepid8 Is there a reason you prefer this approach over using a version manager like e.g. https://github.com/taylor/kiex?

ntrepid8:
@mikemcquaid I wasn't aware of kiex.  It looks like a good solution. Thanks for pointing it out :)

I don't see a kiex formula for homebrew so maybe that would be the best path forward.  I used other tools like this for managing Rust or NodeJS versions and that pattern seems to work pretty well.

mikemcquaid:
> I don't see a kiex formula for homebrew so maybe that would be the best path forward. I used other tools like this for managing Rust or NodeJS versions and that pattern seems to work pretty well.

Yeh, we're trying to encourage version managers where we can. A PR to add support for that or another version manager would be :heart:.

If it's OK I'll close this out in favour of that. Thanks for all your work there and sorry for the recent policy change under you!

ntrepid8:
Yep, let's close this one and I'll take a crack at a formula for kiex :)

PR #1014, opened 2015-11-08
Elasticsearch cleanup and more versions

mikemcquaid:
Mostly wanted to add 1.7 but figured I'd add the in between versions and do some cleanup too.

CC @erebor


DomT4:
I'm mildly unhappy carrying old Elastics here as they tend to be problematic and *nobody* updates them, security problems and all.

If you want them though, won't bite you about it. They need `conflicts_with` though, ideally.

PR #1015, opened 2015-11-08
mysql56 5.6.27 (new formula)

mikemcquaid:
To be merged when https://github.com/Homebrew/homebrew/pull/45670 is.


mikemcquaid:
Failure seems to be a false positive.

PR #1017, opened 2015-11-10
ruby187: cleanup and match core Ruby.

mikemcquaid:
As we may end up using this for Homebrew's testing of older Rubies.


DomT4:
I couldn't get this to build without `:std` env locally a couple months back, FWIW. From the description of the env there in-formula perhaps we could prepend the buildpath to the `$PATH` or something instead, not sure.

mikemcquaid:
Yeh, seems a good idea to fiddle with the buildpath. It must be possible.

mikemcquaid:
Although this is working I don't think it's ready to merge yet. I'm unable to `gem install bundler` due to being unable to require `zlib`.

mikemcquaid:
I'm tempted to merge this as-is and iterate on it. @DomT4 and @xu-cheng if you have any ideas about the `zlib` stuff I'm all :ear:s.

DomT4:
```
/opt/ruby18/bin/irb
irb(main):001:0> require "zlib"
=> true
```
No problems with the one I installed a few weeks ago, so may be a new bug we've exposed. I'll do a rebuild and see what happens.

DomT4:
Ah, yup, with the rebuild as-is I get `LoadError: no such file to load -- zlib`.

DomT4:
The problem goes away if you restore `env :std`. With that in place it doesn't just print `compiling zlib` and then skips over it, which is what is happening with the superenv builds.

```
compiling zlib
/usr/bin/clang -I. -I../.. -I../../. -I../.././ext/zlib -DHAVE_ZLIB_H  -DOS_CODE=OS_UNIX -I/usr/local/opt/openssl/include -I/usr/local/opt/readline/include -F/usr/local/Frameworks -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE  -I/usr/local/opt/openssl/include -I/usr/local/opt/readline/include -F/usr/local/Frameworks -fno-common -Os -w -pipe -march=native -mmacosx-version-min=10.11  -fno-common -pipe -fno-common   -c zlib.c
cc -dynamic -bundle -undefined suppress -flat_namespace -o ../../.ext/i686-darwin15.2.0/zlib.bundle zlib.o -L. -L../.. -L. -L/usr/local/opt/openssl/lib -L/usr/local/opt/readline/lib -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names    -lruby -lz  -ldl -lobjc
```

```
irb(main):001:0> require "zlib"
=> true
```

Wonder what's happening with superenv that makes it skip over the zlib compile stage.

DomT4:
Bunch of logs [here](https://gist.github.com/DomT4/c6c4fd7640a4cb43d8dc). Superenv is breaking *something*, but `¯\_(ツ)_/¯`

UniqMartin:
I think I was able to resolve the remaining issues in PR #1026. (I made a new PR because I wanted the CI to crunch through my changes before declaring success. And it actually caught an issue.)

@mikemcquaid Any reason why you decided to use RubyGems 1.6.2 in this PR? Why not the most recent 2.5.0 which still supports Ruby 1.8.7?

PR #1025, opened 2015-11-16
Adds Docker 1.8.3 version

deiga:


DomT4:
Passing for now as we're trying to cut back on the number of point release we support, but if you believe it meets the new submission [requirements](https://github.com/Homebrew/homebrew-versions#acceptable-formulae) please do feel free to resubmit this formula. Thank you for your submission to Versions! :heart:

deiga:
@DomT4 Do I need to file another PR for this formula or can you just reopen this. I understand the notion for point versions, but in this case this is the last version of `docker` which still works with `boot2docker` so this is kind of needed

DomT4:
> this is the last version of docker which still works with boot2docker so this is kind of needed

I'll take this one, per that, but the rules on this sort of thing are going to get tighter for a fair while. This tap is undergoing a desperately needed cleanup.

deiga:
Nvm, apparently `boot2docker` was updated to use the latest docker :)

DomT4:
Heh, cool. Thanks for checking! :)

PR #1028, opened 2015-11-20
ruby: clean up 1.9.3, 2.0, and 2.1

UniqMartin:
The primary goal was to make the formulae as similar as possible to core `ruby` and the recently merged `ruby187`. Comparing the formulae to each other should now produce a minimal diff.

The most controversial changes might be adding `keg_only` and changing the Gem binary path:

- I feel they should be `keg_only` to allow multiple Rubies from `homebrew/versions` (and core) to be installed in parallel and to stop them from overwriting each other.
- To be consistent with `ruby187`, Gem binary wrappers are now placed in `HOMEBREW_PREFIX/lib/ruby/gems/ABI_VERSION/bin`. This also stops overwriting those wrappers created by core `ruby` which places them in `HOMEBREW_PREFIX/bin`.

Other changes:

- Added descriptions.
- Used HTTPS URLs where appropriate.
- Bumped revisions as there are end-user visible changes.
- Fixed builds with option `--with-suffix`.
- Added required and removed superfluous `configure` arguments and `make` invocations.
- Moved Ruby's `site` and `vendor` directories to `HOMEBREW_PREFIX/lib/ruby`.

I extensively tested option-less, `--with-suffix`, and `--with-doc` builds. I haven't tested `--with-tcltk`.


mikemcquaid:
Thanks @UniqMartin!

PR #1031, opened 2015-11-20
Grails 2.0.1

kflorence:


mikemcquaid:
Why do you need 2.0.1 specifically rather than 2.0.4, out of interest?

kflorence:
@mikemcquaid That is the version we are stuck with for our legacy Grails site, so unfortunately I can't use 2.0.4.

mikemcquaid:
I'm wondering if this will have widespread use as a Homebrew formula or if it's maybe better vendored in projects?

kflorence:
@mikemcquaid does it really matter? Shouldn't homebrew versions support any older version?

mikemcquaid:
CC @DomT4 for thoughts.

DomT4:
Passing under [new guidelines we've agreed and committed](https://github.com/Homebrew/homebrew-versions#acceptable-formulae), which should make future discussions a little clearer. Sorry, and thanks for the submission to Homebrew!

Issue #1027, opened 2015-11-20
Broken bottles


bfontaine:
- [x] cloog018 (all 404)
- [x] gmp4 (all 404)
- [x] isl011 (all 404)
- [x] libmpc08 (all 404)

PR #1032, opened 2015-11-20
isl011: desc & test added

bfontaine:
See #1027.


bfontaine:
@BrewTestBot test this please.

PR #1034, opened 2015-11-21
Added a formula for version 3.9.0 of Watchman

shaheenery:
Certain versions of Ember CLI will fall back to Node Watcher when Watchman is not ~> 3.0.0

### DIFF
--- a/watchman390.rb
+++ b/watchman390.rb
@@ -1,0 +1,40 @@
+class Watchman390 < Formula
+  desc "Watch files and take action when they change"
+  homepage "https://github.com/facebook/watchman"
+  url "https://github.com/facebook/watchman/archive/v3.9.0.tar.gz"
+  sha256 "1739cd2d6846cc688b12911c37406fae5601d76c0d11f3da957c2b7273941221"
+
+  conflicts_with "watchman", :because => "Differing versions of the same formula"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "pcre"
+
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Certain versions of Ember CLI

Which versions? New or old ones?

shaheenery:
Ember CLI 1.13.1 has semver checks that don't recognize watchman 4+ as being supported.  It's fixed in later patch versions, but if you're stuck on 1.13.1 like my team is for now, we would like a way to install this specific version of watchman, with brew, through our provisioning script.

mikemcquaid:
@shaheenery I'd recommend hosting this in your own tap for now. It doesn't feel like this is a formula that will be needed widely or long-term. Sorry and thanks!

shaheenery:
Hey @mikemcquaid , no worries.  I didn't realize that only long-term or widely needed formulas were welcome.  Since it is so easy to have your own custom taps, I guess it really don't matter if a formula makes it into homebrew-versions.  I think I was just assuming that home-brew versions was meant to be a place that held all of the past versions of homebrew formulas that anyone could want, but that *would* make the list incredibly long.  Keep up the good work, thank you for maintaining.

mikemcquaid:
Thanks for the kind words!

DomT4:
We're making some changes to Versions that should end up with some clearer guidelines around this sort of thing. Apologies for the lack of clarity at the moment on what's accepted and what isn't.

shaheenery:
I appreciate the apology, but it's not necessary.  I wanted a formula anyway I'll just add tapping my personal repo to our team's provisioning script.

PR #1036, opened 2015-11-21
antlr2: desc & test added

bfontaine:


bfontaine:
Unfortunately I can’t test locally because I’m over 3G and Antlr2 is too big.

PR #1037, opened 2015-11-21
Gradle*: descs & tests added

bfontaine:


bfontaine:
```
* Exception is:
org.gradle.internal.nativeplatform.NativeIntegrationException: Could not create JNA native library '/Users/brew/.gradle/native/jna/darwin/libjnidispatch.jnilib'.
```

Seems like it’s trying to escape the sandbox.

PR #1043, opened 2015-11-23
gcc49: Add upstream patch to fix assembler generation with XCode 7

TheBloke:
Same issue as was fixed in GCC 5.1, see: https://github.com/Homebrew/homebrew/issues/41633

Upstream bug details: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66509 

Adding the same patch that was used for GCC 5.x (as found in above Issue) works also for GCC 4.9.3, and fixes an assembler-related compilation issue I had using gcc 4.9.3 to compile a project on OSX 10.11.1 with XCode 7.1.1.

According to the GCC bugzilla link, the fix is planned to be backported into GCC 4.9.4.  But in the meantime this patch should fix assembler issues when using GCC 4.9.3 with XCode 7.

### DIFF
--- a/gcc49.rb
+++ b/gcc49.rb
@@ -27,19 +27,18 @@ class Gcc49 < Formula
 
   head "svn://gcc.gnu.org/svn/gcc/branches/gcc-4_9-branch"
 
-  bottle do
-    revision 1
-    sha256 "297b569c6e6992bee96c04c942d2f077665202a9464f9148fc67df358f04edf2" => :el_capitan
-    sha256 "b90f5d13eb7ac46bc4ae1af61c18b2d1889c55a8280af7205a43d73289c7b659" => :yosemite
-    sha256 "37073298a16b73d5c89ef34954cc52df2f3788d400fa6748a25516bac244306e" => :mavericks
-  end
-
   if MacOS.version >= :yosemite
     # Fixes build with Xcode 7.
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66523
     patch do
       url "https://gcc.gnu.org/bugzilla/attachment.cgi?id=35773"
       sha256 "db4966ade190fff4ed39976be8d13e84839098711713eff1d08920d37a58f5ec"
+    end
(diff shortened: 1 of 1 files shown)


gcc49.rb
```diff
@@ -27,13 +27,6 @@ def osmajor
 
   head "svn://gcc.gnu.org/svn/gcc/branches/gcc-4_9-branch"
 
-  bottle do
-    revision 1
-    sha256 "297b569c6e6992bee96c04c942d2f077665202a9464f9148fc67df358f04edf2" => :el_capitan
-    sha256 "b90f5d13eb7ac46bc4ae1af61c18b2d1889c55a8280af7205a43d73289c7b659" => :yosemite
-    sha256 "37073298a16b73d5c89ef34954cc52df2f3788d400fa6748a25516bac244306e" => :mavericks
-  end
```
mikemcquaid: Please leave the bottle block as-is and we'll update it when we pull.

TheBloke: OK so you want me to re-submit this without the bottle removed?

PS I did see that instruction in [Homebrew/Contributing](https://github.com/Homebrew/homebrew/blob/master/CONTRIBUTING.md). However, the instructions in [Homebrew-Versions](https://github.com/Homebrew/homebrew-versions) say:

"Please remove any existing bottle block in your formula prior to submission."

Which I assumed must therefore supersede the general advice and apply for Homebrew-versions requests specifically.

You might want to remove that line if it no longer applies!

mikemcquaid: Yep, I'll be sure that's updated. If you can re-add the removed lines that would be good, thanks.

DomT4: Ping.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/40c61d54b86f9a9a8cec3594a630e53059d53045. Thank you for your contribution to Homebrew @TheBloke; we appreciate it!

PR #1044, opened 2015-11-24
Adds MariaDB 10.0 as main homebrew is now on MariaDB 10.1

kuahyeow:
Based off https://github.com/Homebrew/homebrew/blob/ed4cab6bba4089e555d5e011b2768599c4cf6f1f/Library/Formula/mariadb.rb
with a newer patch level 10.0.22 (rather than 10.0.21)

New sha256sum, of course, removed bottle and devel section as per
homebrew-versions guidelines

Added conflicts with main homebrew "mariadb" recipe

### DIFF
--- a/mariadb100.rb
+++ b/mariadb100.rb
@@ -1,0 +1,208 @@
+class Mariadb100 < Formula
+  desc "Drop-in replacement for MySQL"
+  homepage "https://mariadb.org/"
+  url "https://mirrors.ocf.berkeley.edu/debian/pool/main/m/mariadb-10.0/mariadb-10.0_10.0.22.orig.tar.gz"
+  sha256 "9e17f4aaccb7069a874555cca254974d0f0103f0ad5e2434acab6aa353dafc7b"
+
+  option :universal
+  option "with-tests", "Keep test when installing"
+  option "with-bench", "Keep benchmark app when installing"
+  option "with-embedded", "Build the embedded server"
+  option "with-libedit", "Compile with editline wrapper instead of readline"
+  option "with-archive-storage-engine", "Compile with the ARCHIVE storage engine enabled"
+  option "with-blackhole-storage-engine", "Compile with the BLACKHOLE storage engine enabled"
+  option "with-local-infile", "Build with local infile loading support"
+
+  depends_on "cmake" => :build
+  depends_on "pidof" unless MacOS.version >= :mountain_lion
+  depends_on "openssl"
+
+  conflicts_with "mariadb", :because => "Differing versions of the same formula"
+  conflicts_with "mysql", "mysql-cluster", "percona-server",
+    :because => "mariadb, mysql, and percona install the same binaries."
+  conflicts_with "mysql-connector-c",
+    :because => "both install MySQL client libraries"
+
+  def install
+    # Don't hard-code the libtool path. See:
+    # https://github.com/Homebrew/homebrew/issues/20185
+    inreplace "cmake/libutils.cmake",
+      "COMMAND /usr/bin/libtool -static -o ${TARGET_LOCATION}",
+      "COMMAND libtool -static -o ${TARGET_LOCATION}"
+
+    # Set basedir and ldata so that mysql_install_db can find the server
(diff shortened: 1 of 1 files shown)


mariadb100.rb
```diff
@@ -0,0 +1,226 @@
...
+    if build.with? "tests"
+      (prefix+"mysql-test").cd do
+        system "./mysql-test-run.pl", "status"
+      end
+    else
+      system "mysqld", "--version"
+    end
+  end
+end
+__END__
```
mikemcquaid: Can remove this patch as it's not being used.

kuahyeow:
Thanks @mikemcquaid - Have removed the `__END__` patch. (Didn't know about that before)

mariadb100.rb
```diff
@@ -0,0 +1,210 @@
+class Mariadb100 < Formula
+  desc "Drop-in replacement for MySQL"
+  homepage "https://mariadb.org/"
+  url "http://ftp.osuosl.org/pub/mariadb/mariadb-10.0.22/source/mariadb-10.0.22.tar.gz"
```
dunn: It'd be great to get a `https` url here; looks like Debian has it and you can use the `https://mirrors.kernel.org` mirror: https://packages.debian.org/sid/libmariadbd-dev

kuahyeow: Something like https://mirrors.kernel.org/debian/pool/main/m/mariadb-10.0/mariadb-10.0_10.0.22.orig.tar.gz ? I don't mind. I merely used what mainline homebrew had. Should that be updated too ?

dunn: If you want, sure!

kuahyeow: @dunn Hmm, looks like Debian only has Maria DB 10.0, not 10.1 - I'll leave mainline homebrew alone for now :)

mikemcquaid:
:+1: from me but would like to get @DomT4's thoughts.

mariadb100.rb
```diff
@@ -0,0 +1,210 @@
...
+  option :universal
+  option "with-tests", "Keep test when installing"
+  option "with-bench", "Keep benchmark app when installing"
+  option "with-embedded", "Build the embedded server"
+  option "with-libedit", "Compile with editline wrapper instead of readline"
+  option "with-archive-storage-engine", "Compile with the ARCHIVE storage engine enabled"
+  option "with-blackhole-storage-engine", "Compile with the BLACKHOLE storage engine enabled"
+  option "with-local-infile", "Build with local infile loading support"
+
+  deprecated_option "enable-local-infile" => "with-local-infile"
```
DomT4: Don't need `deprecated_option`s on new formulae, which this essentially is.

mariadb100.rb
```diff
@@ -0,0 +1,210 @@
+class Mariadb100 < Formula
+  desc "Drop-in replacement for MySQL"
+  homepage "https://mariadb.org/"
+  url "https://mirrors.kernel.org/debian/pool/main/m/mariadb-10.0/mariadb-10.0_10.0.22.orig.tar.gz"
```
DomT4: Can you use `https://mirrors.ocf.berkeley.edu/debian/pool/main/` here instead please? The kernel URL has a SHA1 cert chain which looks insecure in Chrome & Chromium these days & whilst not technically a major risk at the moment is an easy enough thing to work around given the abundance of secure Debian mirrors.

DomT4:
Couple of nits, but it's within our new +/-2 range so I don't have a problem with it. We should look at the `conflicts_with` mechanism at some point and decide how comfortable we are with formulae that conflict sharing the same database location given potential data incompatibilities, but this PR is fine as-is.

mariadb100.rb
kuahyeow: Oh, good point about the SHA1. Will do

kuahyeow:
Thanks @DomT4 - have changed to address both your points

mikemcquaid:

For future reference the preferred commit message format for new versions formulae is `mariadb100 10.0.22 (new formula)`.

PR #1047, opened 2015-11-25
openssl102 1.0.2d (new formula)

brandonweeks:
Adds OpenSSL 1.0.2.


mikemcquaid:
Can you provide a bit more context about why this should be added here when it's the same version as in Homebrew core? I don't object, just trying to understand. Also, I think it should probably be called `openssl10` and track `1.0.x` versions.

mikemcquaid:
There's also a few audit failures: http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1045/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_versions_openssl102___strict___online/

DomT4:
Agree with Mike on `openssl10`. There isn't currently a 1.0.3 on the roadmap, but that doesn't make it an impossibility.

FWIW; if upstream's timetable doesn't move, I expect Homebrew to move onto OpenSSL 1.1.0 sometime in early to the middle of May. It will likely take us at least one or two weeks to resolve any incompatibilities caused by the intentional 1.1.0 API breakage, and most to all of our OpenSSL-using binary packages are expected to need rebuilding.

PR #1048, opened 2015-11-28
gradle29: added

mosampaio:
adds gradle 2.9


mikemcquaid:
We're reconsidering some of the new things in homebrew-versions. What's your use-case for needing an older version? ThankS!

PR #539, opened 2014-10-08
Sphinx 2.0.8

shosti:
New versions aren't completely backward-compatible.

PR #1050, opened 2015-11-30
add sphinx 2.0.8

chaimleib:
# Overview
Newer versions of [sphinx](http://www.sphinxsearch.com) have a different API, which breaks compatibility with software dependent on it.

The test is basic, because a more thorough one would require installing at least `mysql` or `postgresql` as the backing store, plus a client library to actually query sphinx. (Note that the included `search` tool was not intended for production use and is [known-bad](http://sphinxsearch.com/bugs/view.php?id=1424) in this version of sphinx.)

# Thanks and reference
This is a modified version of PR #539. Thanks to @shosti for the [original code](https://github.com/shosti/homebrew-versions/blob/b7c09538a91a7dc13937be26a94067de5e05711f/sphinx208.rb).
A copy of this formula is in https://github.com/chaimleib/homebrew-formulae, where it has been tested by many users.

### DIFF
--- a/sphinx208.rb
+++ b/sphinx208.rb
@@ -1,0 +1,48 @@
+class Sphinx208 < Formula
+  desc "Full-text search engine server"
+  homepage "http://www.sphinxsearch.com"
+  url "http://sphinxsearch.com/files/sphinx-2.0.8-release.tar.gz"
+  sha256 "bd699bf34635558e4d9d4d8bdcf5fcdd9c2e8f6c614824bec793b0010fa6cda3"
+
+  option "with-mysql", "Force compiling against MySQL"
+  option "with-postgresql", "Force compiling against PostgreSQL"
+
+  depends_on :mysql => :optional
+  depends_on :postgresql => :optional
+
+  conflicts_with "sphinx", :because => "differing version of same formula"
+
+  def install
+    args = %W[--prefix=#{prefix}
+              --disable-dependency-tracking
+              --localstatedir=#{var}]
+
+    args << "--#{build.with?("mysql") ? "with" : "without"}-mysql"
+    args << "--#{build.with?("postgresql") ? "with" : "without"}-pgsql"
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  def caveats; <<-EOS.undent
+    Sphinx depends on either MySQL or PostreSQL as a datasource.
+
+    You can install these with Homebrew with:
+      brew install mysql
+        For MySQL server.
+
+      brew install mysql-connector-c
+        For MySQL client libraries only.
+
+      brew install postgresql
+        For PostgreSQL server.
+
+    We don't install these for you when you install this formula, as
+    we don't know which datasource you intend to use.
+    EOS
+  end
+
+  test do
+    assert_match /#{version}/, shell_output("#{bin}/searchd -h")
+  end
+end


sphinx208.rb
```diff
@@ -0,0 +1,51 @@
+class Sphinx208 < Formula
+  desc "Full-text search engine server"
+  homepage "http://www.sphinxsearch.com"
+  url "http://sphinxsearch.com/files/sphinx-2.0.8-release.tar.gz"
+  sha256 "bd699bf34635558e4d9d4d8bdcf5fcdd9c2e8f6c614824bec793b0010fa6cda3"
+
+  option "with-mysql", "Force compiling against MySQL"
+  option "with-postgresql", "Force compiling against PostgreSQL"
+
+  deprecated_option "mysql" => "with-mysql"
```
DomT4: Can't deprecate stuff in a new formula :)

sphinx208.rb
```diff
@@ -0,0 +1,51 @@
...
+  deprecated_option "pgsql" => "with-postgresql"
+
+  depends_on :mysql => :optional
+  depends_on :postgresql => :optional
+
+  conflicts_with "sphinx"
```
DomT4: Want a `:because =>` explanation here.

sphinx208.rb
chaimleib: :) In the [original version](https://github.com/Homebrew/homebrew-versions/pull/539/files) of this script, these options were provided. So it's not exactly a new formula.

sphinx208.rb
chaimleib: Anything special to mention here @shosti?

sphinx208.rb
DomT4: It is a new formula though, technically. It has a new name and nobody will be installing with the existing wrongly-named options.

chaimleib:
jenkins test again?

chaimleib:
I think I've fixed all the issues. Is there anything else to do?

mikemcquaid:
> Newer versions of sphinx have a different API, which breaks compatibility with software dependent on it.

What software is that? Can you provide some examples? Also, @DomT4 may want to express views on including this as we're being much more selective about what we include in `homebrew-versions` now. Thanks!

chaimleib:
The software in question is the [Coupa](http://www.coupa.com/) web app, and the associated Coupa Supplier Network. In the interest of stability, we need access to the older version of sphinx search.

mikemcquaid:
> These are large Rails projects for Enterprise, and in the interest of stability, we need access to the older version of sphinx search.

Can you clarify what you mean by "in the interest of stability"? If it's just "we'd like to pin the version" then I'd recommend you create your own tap for this.

chaimleib:
Yes, I meant API stability. I thought this is why everyone tapped homebrew-versions. Could you explain what the new criteria for inclusion are?

mikemcquaid:
https://github.com/Homebrew/homebrew-versions/blob/master/README.md#acceptable-formulae is the current guidance but we're trying to generally be more conservative here. @DomT4 has the final word, though.

DomT4:
I don't have a massive problem with inclusion on the basis of API breaks, but I'm curious why you specifically want us to host it when you've already set it up in a tap yourself? Does it take particularly long to build, or?

chaimleib:
Mostly I wanted to contribute something that worked for me and simplify installation for other users. It takes about a minute to build, so it would benefit from bottling, although I'm not sure whether bottling works with compile-time options for PostgreSQL or for MySQL.

mikemcquaid:
Let's leave this open and see if we get any other :+1:s. Thanks either way @chaimleib!

DomT4:
Going to pass here because there hasn't been any additional interest and I'd prefer to track branches rather than patch releases going forwards; the latest release here would be 2.0.9 but even that was back in 2013.

Thank you nonetheless for the submission to Homebrew.

PR #976, opened 2015-09-30
elasticsearch103: New Formula

chiubaka:
Needed elasticsearch 1.0.3 and couldn't find it on homebrew or homebrew-versions.

PR #1051, opened 2015-12-01
elasticsearch103: New Formula

chiubaka:
Needed elasticsearch 1.0.3 and couldn't find it on homebrew or homebrew-versions.

Follow-up on pull request #976, but this time I've run `brew audit --strict` and verified that it comes out clean.


mikemcquaid:
I'm afraid we're no longer accepting really old versions in here any more: https://github.com/Homebrew/homebrew-versions#acceptable-formulae

Sorry!

mikemcquaid:
Please consider [creating your own tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) instead. Thanks!

PR #1061, opened 2015-12-04
Add elasticsearch 1.2.1

rwwagner90:
We needed elasticsearch 1.2.1 rather than 1.2.4 that is included here, so I made elasticsearch121.rb. I think other people probably need it too, since there is an article talking about how to do it http://effectif.com/mac-os-x/installing-specific-version-of-homebrew-formula

The article is out of date, and no longer works, so we need something here for others that require elasticsearch 1.2.1


mikemcquaid:
Why do you need 1.2.1 rather than 1.2.4?

rwwagner90:
@mikemcquaid I'm not entirely sure why we need 1.2.1 over 1.2.4, but my back end guy told me we need 1.2.1 specifically. Something about the way locations are stored and location searches I think.

mikemcquaid:
@rwwagner90 We don't want to track patch versions like 1.2.1 rather than 1.2.4; there should only be security/bugfix differences rather than API differences. FWIW running ElasticSearch on OS X (I'm assuming your production is on Linux) will likely have bigger differences than running 1.2.4 vs 1.2.1. Sorry and thanks for the PR!

PR #1064, opened 2015-12-07
Add in a specific reference to grails 2.1.5

khanage:
There is a grails21 already (which this is just a copy paste of). I've created something specific rather than updating the existing grails21 so as not to break anything for anyone else.

### DIFF
--- a/grails215.rb
+++ b/grails215.rb
@@ -1,0 +1,28 @@
+class Grails215 < Formula
+  desc "Web application framework for the Groovy language"
+  homepage "http://grails.org"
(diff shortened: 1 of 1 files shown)


DomT4:
Can you justify including this for a specific reason, beyond it might break? How wide is usage going to be?

We're trying to stop tracking point releases as part of the new [submission rules](https://github.com/Homebrew/homebrew-versions#acceptable-formulae), because compared to the core 99% of all contributions and fixes in Versions come from maintainers and it's an unreasonably heavy burden on the project for a tap that *relatively* few people use.

PR #1068, opened 2015-12-09
Remove unmaintained Elasticsearch versions

jasontedor:
This commit removes formulae for versions of Elasticsearch that are no
longer actively maintained.


DomT4:
I couldn't be happier to see all of these go to the big ol' formula graveyard in the sky, but @mikemcquaid may have comment. He did a big cleanup of Elastic stuff in Versions recently.

mikemcquaid:
I'm :+1: on these being :skull:d (selfishly because we don't use these versions). @DomT4 thoughts on whether these should :skull: or move to `homebrew-versions-unsupported`? or `homebrew-versions-boneyard`?

DomT4:
Anything that has an unpatched security hole that has no chance of being patched can :skull: as far as I'm concerned. Anything else can live in `homebrew-versions-unsupported`, which I'll actually go create shortly.

mikemcquaid:
@DomT4 I guess my thinking is: I'm sure those all have similar security problems, even if we don't know about them.

mikemcquaid:
@DomT4 (which isn't an argument against merging this but I think is perhaps an argument against `homebrew-versions-unsupported`

jasontedor:
> Anything that has an unpatched security hole that has no chance of being patched

[Versions < 1.6.1 have known issues](https://www.elastic.co/community/security) that will not be patched, and any unknown issues will also not be patched. Only versions 1.7.x, 2.0.x, and 2.1.x are currently being maintained.

That applies to all of the formulas removed in this pull request.

DomT4:
> which isn't an argument against merging this but I think is perhaps an argument against homebrew-versions-unsupported

I'm willing to draw a narrow distinction between stuff that has security problems and stuff that has security problems that may be intended to be public facing. I appreciate how narrow a distinction this is.

mikemcquaid:
> I'm willing to draw a narrow distinction between stuff that has security problems and stuff that has security problems that may be intended to be public facing. I appreciate how narrow a distinction this is.

That's a reasonable distinction :+1:. We'll be much harsher on servers than libraries/clients. Sounds good.

PR #1077, opened 2015-12-13
webkitgtk 2.4.9 (new formula)

eallik:
`webkitgtk-2.4.9` is the last version of webkitgtk to support the WebKit1 API; this is important for libraries that depend on version 1 of the API, such as [webkitgtk3](https://hackage.haskell.org/package/webkitgtk3).

So here is a formula for 2.4.9 (tested on Mavericks). It compiles against Quartz (not X11, but X11 support could be added — the MacPorts Portfile has support for both, and possibly even Wayland I think).

It support WebGL (but seems to link against `/opt/X11` not `OpenGL.framework`), geo-location, web video, GObject Introspection, SVG, but does not support WebKit2, web-audio, accelerated compositing nor JIT compilation.

It prefers non-Apple gnumake but can also compile with Apple's (`/usr/bin/make`) but will then `deparallelize` the `ENV` to work around hangs.

It currently works in `std` by default but can easily build in superenv as well, but will then, again, `deparallelize` to avoid hangs. If anyone could help me fix that, I'd really appreciate that.

The formula was based on the Portfile (also mentioned in the commets); (almost) all of the patches from the original port were copied over, with slight adjustments/updates.

The unsupported `configure` flags were left in and exposed, just to draw attention of anyone who might need them and perhaps even be able to help add support for them.

### DIFF
--- a/webkitgtk24.rb
+++ b/webkitgtk24.rb
@@ -70,10 +70,10 @@ class Webkitgtk24 < Formula
       --without-gtk-doc
     ]
 
-    args.push flag_for("video"),
-              flag_for("webgl"),
-              flag_for("homebrew/x11/geoclue", "geolocation"),
-              flag_for("gobject-introspection", "introspection")
+    args << build.with?("video") ? "--enable-video" : "--disable-video"
+    args << build.with?("webgl") ? "--enable-webgl" : "--disable-webgl"
+    args << build.with?("homebrew/x11/geoclue") ? "--enable-geolocation" : "--disable-geolocation"
+    args << build.with?("gobject-introspection") ? "--enable-geolocation" : "--disable-geolocation"
 
     # from the original Portfile
     ENV.append ["CFLAGS", "CXXFLAGS"], "-ftemplate-depth=256"
@@ -83,8 +83,6 @@ class Webkitgtk24 < Formula
     # from https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
     ENV.append "CXXFLAGS", "-std=gnu++11"
 
-    # this will make the build much much slower, but if you remove it,
-    # you most likely also have to use non-Apple gnumake and env :std
     ENV.deparallelize
 
     system "autoreconf", "-fvi"
--- a/webkitgtk24.rb
+++ b/webkitgtk24.rb
@@ -1,62 +1,22 @@
-USE_STD_ENV = true
-
 class Webkitgtk24 < Formula
-  # This formula is ported from (or based on) the following MacPorts port:
-  # https://svn.macports.org/repository/macports/trunk/dports/www/webkit-gtk/Portfile
-
-  # also relevant:
-  # https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
-
   desc "Full-featured Gtk+ port of the WebKit rendering engine"
   homepage "http://webkitgtk.org"
   url "http://webkitgtk.org/releases/webkitgtk-2.4.9.tar.xz"
   sha256 "afdf29e7828816cad0be2604cf19421e96d96bf493987328ffc8813bb20ac564"
 
-  conflicts_with "webkitgtk"
-
   needs :cxx11
 
-  # TODO: how can we switch back to superenv while not making parallel make hang?
-  env :std if USE_STD_ENV
+  option "without-video", "disable HTML5 video object support"
+  option "without-webgl", "disable WebGL support"
 
-  option "without-geolocation", "DISABLE geolocation support"
-  option "without-video", "DISABLE HTML5 video object support"
-  option "without-introspection", "DISABLE GObject introspection"
-  option "without-webgl", "DISABLE WebGL support"
-
-  # the original Portfile says (and it seems to hold):
-  #
-  # > Apple's gnumake (patched 3.81) gets wedged at some point during the build process
-  #
-  # so we make this optional and disable parallelism if Apple's make is used
-  option "with-apple-gnumake", "use Apple's gnumake at /usr/bin/make"
-
-  # let's still keep these to draw attention
-  option "without-svg", "(not supported) DISABLE SVG support"
-  option "with-gtk-doc", "(not supported) ENABLE documentation generation"
-  option "with-accelerated-compositing", "(not supported) ENABLE accelerated compositing support"
-  option "with-jit", "(not supported) ENABLE JIT compilation support"
-  option "with-web-audio", "(not supported) ENABLE web-audio support"
-  option "with-webkit2", "(not supported) ENABLE WebKit 2 support"
-
-  bottle do
-    sha256 "a1d3ad20a44da0a303f957a2259be0311a730e94c2f3d716d2f9c300c173adbe" => :el_capitan
-    sha256 "500f31ae4c8ebbc19d0a85a8bd2573e8102804ec7f7d9e172e3a01ea689a0865" => :yosemite
-    sha256 "0c66a6ad12c4ced2741332a0e6e60510ef2e0a9f739597e9fb2b5117e31454ea" => :mavericks
-  end
-
-  depends_on "make"       => :build if build.without? "apple-gnumake"
-  depends_on "autoconf"   => :build
-  depends_on "automake"   => :build
-  depends_on "libtool"    => :build
-  depends_on "perl"       => :build
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "perl" => :build
   depends_on "pkg-config" => :build
-  depends_on "python"     => :build
+  depends_on "python" => :build if MacOS.version <= :snow_leopard
 
   depends_on "glib"
   depends_on "gtk+3"
-  depends_on "bison"
-  depends_on "flex"
   depends_on "enchant"
   depends_on "libffi"
   depends_on "harfbuzz"
@@ -68,112 +28,18 @@ class Webkitgtk24 < Formula
   depends_on "sqlite"
   depends_on "webp"
 
-  depends_on "geoclue"               if build.with? "geolocation"
-  depends_on "gobject-introspection" if build.with? "introspection"
+  depends_on "geoclue" => :recommended
+  depends_on "gobject-introspection" => :recommended
 
-  # TODO: not sure if web-audio also requires gstreamer & gst-plugins-base
-  if (build.with? "web-audio") || (build.with? "video")
+  if build.with? "video"
     depends_on "gstreamer"
     depends_on "gst-plugins-base"
   end
 
-  # https://lists.webkit.org/pipermail/webkit-gtk/2013-November/001630.html
-  depends_on "gtk+" if build.with? "webkit2"
-
-  depends_on "gtk-doc" => :build if build.with? "gtk-doc"
-
-  ### BEGIN BROKEN FEATURE WARNINGS
-
-  # see https://bugs.webkit.org/show_bug.cgi?id=126438
-  #     https://bugs.webkit.org/show_bug.cgi?id=126437
-  opoo "web-audio support is currently broken; see the formula for details"               if build.with? "web-audio"
-
-  # see https://bugs.webkit.org/show_bug.cgi?id=133293
-  opoo "JIT compilation support is currently broken; see the formula for details"         if build.with? "jit"
-
-  # see https://trac.macports.org/ticket/41663
-  opoo "accelerated compositing support is currently broken; see the formula for details" if build.with? "accelerated-compositing"
-
-  # see https://lists.webkit.org/pipermail/webkit-gtk/2013-November/001629.html
-  #
-  # ATTENTION: enable this by default, if possible, but then also see the
-  # corresponding patch block below.
-  #
-  opoo "WebKit 2 support is currently broken; see the formula for details"                if build.with? "webkit2"
-
-  # disabling SVG appears to fail the build
-  #    "error: use of undeclared identifier 'CSSValueAlphabetic'"
-  #
-  # see https://lists.webkit.org/pipermail/webkit-dev/2014-January/026166.html
-  opoo "building without SVG support is currently broken; see the formula for details"    if build.without? "svg"
-
-  # odie "GObject introspection support is currently broken; see the formula for details"   if     build.with? "introspection"
-
-  ### END BROKEN FEATURE WARNINGS
-
+  # see the inline comments
   patch :DATA
 
   def install
-    make_command = (build.with? "apple-gnumake") ? "make" : "gmake"
-
-    do_replaces
-
-    @configure_args = [
-      "--prefix=#{prefix}",
-
-      # see https://lists.webkit.org/pipermail/webkit-unassigned/2014-November/645071.html
-      "--enable-dependency-tracking",
-
-      # https://bugs.webkit.org/show_bug.cgi?id=94488
-      # "--disable-dependency-tracking",
-
-      "--enable-x11-target=no",
-      "--enable-quartz-target=yes",
-      "--enable-wayland-target=no",
-    ]
-
-    flags = ["web-audio", "video", "webgl", "geolocation", "svg", "jit",
-             "accelerated-compositing", "webkit2", "gtk-doc", "introspection"]
-    flags.each do |opt|
-      enable_or_disable = (build.with? opt) ? "enable" : "disable"
-      @configure_args.push "--#{enable_or_disable}-#{opt}"
-    end
-
-    # from the original Portfile
-    ENV.append ["CFLAGS", "CXXFLAGS"], "-ftemplate-depth=256"
-    ENV.append "CXXFLAGS", "-Wno-c++11-extensions"
-    ENV.append "CPPFLAGS", "-DGTEST_USE_OWN_TR1_TUPLE=1"
-
-    # from https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
-    ENV.append "CXXFLAGS", "-std=gnu++11"
-
-    if (build.with? "apple-gnumake") || !USE_STD_ENV
-      # (see also http://stackoverflow.com/a/34206868/247623)
-      #
-      # without `env :std` and without `ENV.deparallelize`, make will stall at:
-      #
-      #     libtool: link: (cd ... && rm -rf ...)
-      #
-      opoo "non-parallelized builds can be very slow"
-
-      ENV.deparallelize
-      #
-      # consider also (but this seemed to offer no help for me):
-      #
-      # ENV['LIPO'] = "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/lipo"
-    end
-
-    system "autoreconf", "-fvi"
-
-    system "./configure", *@configure_args
-
-    system make_command, "-w", "all", "V=1"
-    system make_command, "install-am"
-    system make_command, "all-am"
-    system make_command, "WebKitGTK-3.0.pot-update"
-  end
-
-  def do_replaces
     # https://bugs.webkit.org/show_bug.cgi?id=126433
     inreplace "Source/JavaScriptCore/API/WebKitAvailability.h",
               "defined(__APPLE__)",
@@ -185,23 +51,52 @@ class Webkitgtk24 < Formula
       "#define PLATFORM(MAC)",
     ]
 
-    # https://bugs.webkit.org/show_bug.cgi?id=126325
-    #
-    # original reinplace call in Portfile is:
-    #     reinplace "s:-stdlib=libstdc\+\+::" ${worksrcpath}/Source/autotools/SetupCompilerFlags.m4
-    #
-    # ...but the actual line in that file is:
-    #     AS_CASE([$CXXFLAGS], [*-stdlib=*], [], [CXXFLAGS="$CXXFLAGS -stdlib=libc++"])
-    #
-    # so let's try this instead:
-    # inreplace "Source/autotools/SetupCompilerFlags.m4",
-    #           "-stdlib=libc++",
-    #           ""
-
     # https://bugs.webkit.org/show_bug.cgi?id=126329
     inreplace "Source/JavaScriptCore/API/JSBase.h",
               /^#define JSC_OBJC_API_ENABLED.*$/,
               "#define JSC_OBJC_API_ENABLED 0"
+
+    args = %W[
+      --prefix=#{prefix}
+
+      --enable-dependency-tracking
+
+      --enable-x11-target=no
+      --enable-quartz-target=yes
+      --enable-wayland-target=no
+      --disable-accelerated-compositing
+      --disable-jit
+      --with-svg
+      --without-web-audio
+      --without-webkit2
+      --without-gtk-doc
+    ]
+
+    args.push flag_for("video"),
+              flag_for("webgl"),
+              flag_for("geoclue", "geolocation"),
+              flag_for("gobject-introspection", "introspection")
+
+    # from the original Portfile
+    ENV.append ["CFLAGS", "CXXFLAGS"], "-ftemplate-depth=256"
+    ENV.append "CXXFLAGS", "-Wno-c++11-extensions"
+    ENV.append "CPPFLAGS", "-DGTEST_USE_OWN_TR1_TUPLE=1"
+
+    # from https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
+    ENV.append "CXXFLAGS", "-std=gnu++11"
+
+    # this will make the build much much slower, but if you remove it,
+    # you most likely also have to use non-Apple gnumake and env :std
+    ENV.deparallelize
+
+    system "autoreconf", "-fvi"
+
+    system "./configure", *args
+
+    system "make", "-w", "all", "V=1"
+    system "make", "install-am"
+    system "make", "all-am"
+    system "make", "WebKitGTK-3.0.pot-update"
   end
 
   def prepend_lines(fname, lines)
@@ -209,6 +104,10 @@ class Webkitgtk24 < Formula
     lines_str = lines.map { |x| x + "\n" }.join
     new = lines_str + old
     File.open(fname, "w") { |f| f.puts new }
+  end
+
+  def flag_for(opt, flag_name = opt)
+    "--#{(build.with? opt) ? "enable" : "disable"}-#{flag_name}"
   end
 end
 
--- a/webkitgtk24.rb
+++ b/webkitgtk24.rb
@@ -14,6 +14,7 @@ class Webkitgtk24 < Formula
   depends_on "perl" => :build
   depends_on "pkg-config" => :build
   depends_on "python" => :build if MacOS.version <= :snow_leopard
+  depends_on "libtool" => :build  # for glibtoolize
 
   depends_on "glib"
   depends_on "gtk+3"
--- a/webkitgtk24.rb
+++ b/webkitgtk24.rb
@@ -28,7 +28,7 @@ class Webkitgtk24 < Formula
   depends_on "sqlite"
   depends_on "webp"
 
-  depends_on "geoclue" => :recommended
+  depends_on "homebrew/x11/geoclue" => :recommended
   depends_on "gobject-introspection" => :recommended
 
   if build.with? "video"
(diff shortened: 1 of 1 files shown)


WebReflection:
since there's still no way to star a PR on Github, I'll disturbingly +1 this, hoping  a formula for WebKi2GTK3 would make it too so I can  forget about MacPorts which is incompatible with pretty much everything else.

Thanks!

eallik:
@WebReflection, I'm curious to know what do you mean exactly by "WebKit2GTK3"? There is already webkitgtk-2.10.4 formula available in Homebrew.

WebReflection:
@eallik is that WebKit2 API for GTK3 ? If so, I'm trying it right now, not sure it worked before but thanks for the info.

eallik:
I'm pretty sure webkitgtk-2.10.4 supports both the WebKit2 API as well as GTK3 or GTK+3 or smth (not at all familiar with GTK tbh), but I'm not sure if that qualifies as "WebKIt2 API for GTK3". I'd be interested to find out though, so if you could keep us posted...

eallik:
By the way, I'm not sure how to address the "all checks have failed" message — `brew audit --strict --online webkitgtk24` passed with 0 complaints, but the formula itself has no `def test` because I couldn't come up with one (the Portfile didn't have one and the one in `webkitgtk-2.10.4` does not work out of the box on this one).

So I'm hoping it's OK to leave this as is.

WebReflection:
I surely need to dig more on this, but basically this is the error that I
don't have with the MacPort version.

  1. it is WebKit2 indeed, my bad I haven't checked before
  2. it is loaded through some binding and fails

```
** (gjs:18901): WARNING **: Failed to load shared library
'/tmp/webkitgtk20151207-6932-1shmxi0/webkitgtk-2.10.4/lib/libwebkit2gtk-4.0.37.dylib'
referenced by the typelib:
dlopen(/tmp/webkitgtk20151207-6932-1shmxi0/webkitgtk-2.10.4/lib/libwebkit2gtk-4.0.37.dylib,
9): image not found


** (gjs:18901): WARNING **: Failed to load shared library
'/tmp/webkitgtk20151207-6932-1shmxi0/webkitgtk-2.10.4/lib/libjavascriptcoregtk-4.0.18.dylib'
referenced by the typelib:
dlopen(/tmp/webkitgtk20151207-6932-1shmxi0/webkitgtk-2.10.4/lib/libjavascriptcoregtk-4.0.18.dylib,
9): image not found


(gjs:18901): Gjs-WARNING **: JS ERROR: Error: Unsupported type void,
deriving from fundamental void

```

Will eventually post or keep updated somewhere else since this is not about
current PR (but I'll check that too)


Regards

eallik:
I'm also not sure if the `conflicts_with` declaration is needed — since the two formulas have a different name, and they have nothing under `bin`, and they provide different versions of the `webkitgtk` package (`pkg-config`), they should be able to live happily side by side with one another...?

dunn:
We're really trying to avoid `std`.  What's the benefit here?

DomT4:
Will probably flat-out reject any new formula that uses it, to be honest.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
+USE_STD_ENV = true
+
+class Webkitgtk24 < Formula
+  # This formula is ported from (or based on) the following MacPorts port:
```
DomT4: Don't need the explanation.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+  # also relevant:
+  # https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
+
+  desc "Full-featured Gtk+ port of the WebKit rendering engine"
+  homepage "http://webkitgtk.org"
+  url "http://webkitgtk.org/releases/webkitgtk-2.4.9.tar.xz"
+  sha256 "afdf29e7828816cad0be2604cf19421e96d96bf493987328ffc8813bb20ac564"
+
+  conflicts_with "webkitgtk"
```
DomT4: Please add a `:because =>`

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+  needs :cxx11
+
+  # TODO: how can we switch back to superenv while not making parallel make hang?
```
DomT4: Use `ENV.deparallelize` if necessary. We're not using `env :std` in new formulae.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  env :std if USE_STD_ENV
+
+  option "without-geolocation", "DISABLE geolocation support"
```
DomT4: Don't block-caps `DISABLE`.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  option "without-video", "DISABLE HTML5 video object support"
+  option "without-introspection", "DISABLE GObject introspection"
+  option "without-webgl", "DISABLE WebGL support"
+
+  # the original Portfile says (and it seems to hold):
+  #
+  # > Apple's gnumake (patched 3.81) gets wedged at some point during the build process
+  #
+  # so we make this optional and disable parallelism if Apple's make is used
+  option "with-apple-gnumake", "use Apple's gnumake at /usr/bin/make"
```
DomT4: We use Apple's `make` unconditionally in Homebrew.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+  # let's still keep these to draw attention
+  option "without-svg", "(not supported) DISABLE SVG support"
```
DomT4: Again on the block caps.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  option "with-gtk-doc", "(not supported) ENABLE documentation generation"
+  option "with-accelerated-compositing", "(not supported) ENABLE accelerated compositing support"
+  option "with-jit", "(not supported) ENABLE JIT compilation support"
+  option "with-web-audio", "(not supported) ENABLE web-audio support"
+  option "with-webkit2", "(not supported) ENABLE WebKit 2 support"
+
+  bottle do
```
DomT4: We'll add this block, you don't have to.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+    sha256 "a1d3ad20a44da0a303f957a2259be0311a730e94c2f3d716d2f9c300c173adbe" => :el_capitan
+    sha256 "500f31ae4c8ebbc19d0a85a8bd2573e8102804ec7f7d9e172e3a01ea689a0865" => :yosemite
+    sha256 "0c66a6ad12c4ced2741332a0e6e60510ef2e0a9f739597e9fb2b5117e31454ea" => :mavericks
+  end
+
+  depends_on "make"       => :build if build.without? "apple-gnumake"
```
DomT4: See notation above on the `make` issue.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  depends_on "autoconf"   => :build
```
DomT4: Let's not bother adding in redundant spacing here. Lining the `:build` lines up is unimportant, and a lot of editors will highlight the extra unnecessary spacing these days.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  depends_on "harfbuzz"
+  depends_on "icu4c"
+  depends_on "libxslt"
+  depends_on "libpng"
+  depends_on "libsecret"
+  depends_on "libsoup"
+  depends_on "sqlite"
+  depends_on "webp"
+
+  depends_on "geoclue"               if build.with? "geolocation"
```
DomT4: Same comment on spacing.

DomT4: Also we're trying to avoid `if build.with?` dependencies outside `def install`, so ideally the option would be `"with-geoclue", "Enables geolocation support"` and then `depends_on "geoclue" => :optional` or similar.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  depends_on "gobject-introspection" if build.with? "introspection"
```
DomT4: Same note here as above.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+  # TODO: not sure if web-audio also requires gstreamer & gst-plugins-base
```
DomT4: Please check this :)

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  # so we make this optional and disable parallelism if Apple's make is used
+  option "with-apple-gnumake", "use Apple's gnumake at /usr/bin/make"
+
+  # let's still keep these to draw attention
+  option "without-svg", "(not supported) DISABLE SVG support"
+  option "with-gtk-doc", "(not supported) ENABLE documentation generation"
+  option "with-accelerated-compositing", "(not supported) ENABLE accelerated compositing support"
+  option "with-jit", "(not supported) ENABLE JIT compilation support"
+  option "with-web-audio", "(not supported) ENABLE web-audio support"
+  option "with-webkit2", "(not supported) ENABLE WebKit 2 support"
```
DomT4: I'd prefer to drop unsupported options initially. We can consider adding them back if people request them, but more options = more breakage we never notice until users report it and we have to spend hours debugging it.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+    sha256 "0c66a6ad12c4ced2741332a0e6e60510ef2e0a9f739597e9fb2b5117e31454ea" => :mavericks
+  end
+
+  depends_on "make"       => :build if build.without? "apple-gnumake"
+  depends_on "autoconf"   => :build
+  depends_on "automake"   => :build
+  depends_on "libtool"    => :build
+  depends_on "perl"       => :build
+  depends_on "pkg-config" => :build
+  depends_on "python"     => :build
```
DomT4: Should be `depends_on :python => :build if MacOS.version <= :snow_leopard`

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+    depends_on "gstreamer"
+    depends_on "gst-plugins-base"
+  end
+
+  # https://lists.webkit.org/pipermail/webkit-gtk/2013-November/001630.html
+  depends_on "gtk+" if build.with? "webkit2"
+
+  depends_on "gtk-doc" => :build if build.with? "gtk-doc"
+
+  ### BEGIN BROKEN FEATURE WARNINGS
```
DomT4: Yeah, this is pretty much why I don't want broken features compiled in. We will end up supporting them if we offer them, even if upstream have decided not to.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+  #    "error: use of undeclared identifier 'CSSValueAlphabetic'"
+  #
+  # see https://lists.webkit.org/pipermail/webkit-dev/2014-January/026166.html
+  opoo "building without SVG support is currently broken; see the formula for details"    if build.without? "svg"
+
+  # odie "GObject introspection support is currently broken; see the formula for details"   if     build.with? "introspection"
+
+  ### END BROKEN FEATURE WARNINGS
+
+  patch :DATA
```
DomT4: Where is this patch from/what is it doing/has it been reported upstream/where's the upstream bug submission link/etc. Thanks!

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+  def install
+    make_command = (build.with? "apple-gnumake") ? "make" : "gmake"
```
DomT4: Same notation as further up on make.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+    do_replaces
+
+    @configure_args = [
```
DomT4: Just:

```ruby
args = %W[
  --prefix=#{prefix}
  --enable-x11-target=no
  --enable-quartz-target=yes
]
```
And so on.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+      # "--disable-dependency-tracking",
+
+      "--enable-x11-target=no",
+      "--enable-quartz-target=yes",
+      "--enable-wayland-target=no",
+    ]
+
+    flags = ["web-audio", "video", "webgl", "geolocation", "svg", "jit",
+             "accelerated-compositing", "webkit2", "gtk-doc", "introspection"]
+    flags.each do |opt|
```
DomT4: For the sake of readability, would prefer these reside on their own lines.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+      enable_or_disable = (build.with? opt) ? "enable" : "disable"
+      @configure_args.push "--#{enable_or_disable}-#{opt}"
+    end
+
+    # from the original Portfile
```
DomT4: Would be good to get some kind of upstream references here.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+    ENV.append ["CFLAGS", "CXXFLAGS"], "-ftemplate-depth=256"
+    ENV.append "CXXFLAGS", "-Wno-c++11-extensions"
+    ENV.append "CPPFLAGS", "-DGTEST_USE_OWN_TR1_TUPLE=1"
+
+    # from https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
+    ENV.append "CXXFLAGS", "-std=gnu++11"
+
+    if (build.with? "apple-gnumake") || !USE_STD_ENV
```
DomT4: Same notations about make and `env :std`.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+      ENV.deparallelize
+      #
+      # consider also (but this seemed to offer no help for me):
+      #
+      # ENV['LIPO'] = "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/lipo"
+    end
+
+    system "autoreconf", "-fvi"
+
+    system "./configure", *@configure_args
```
DomT4: `system "./configure", *args`

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+
+    system make_command, "-w", "all", "V=1"
+    system make_command, "install-am"
+    system make_command, "all-am"
+    system make_command, "WebKitGTK-3.0.pot-update"
+  end
+
+  def do_replaces
```
DomT4: These should be inlined into `def install`, Thanks.

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+    sha256 "a1d3ad20a44da0a303f957a2259be0311a730e94c2f3d716d2f9c300c173adbe" => :el_capitan
+    sha256 "500f31ae4c8ebbc19d0a85a8bd2573e8102804ec7f7d9e172e3a01ea689a0865" => :yosemite
+    sha256 "0c66a6ad12c4ced2741332a0e6e60510ef2e0a9f739597e9fb2b5117e31454ea" => :mavericks
+  end
+
+  depends_on "make"       => :build if build.without? "apple-gnumake"
+  depends_on "autoconf"   => :build
+  depends_on "automake"   => :build
+  depends_on "libtool"    => :build
+  depends_on "perl"       => :build
```
dunn: The system Perl can't be used?

webkitgtk24.rb
```diff
@@ -0,0 +1,648 @@
...
+      opoo "non-parallelized builds can be very slow"
+
+      ENV.deparallelize
+      #
+      # consider also (but this seemed to offer no help for me):
+      #
+      # ENV['LIPO'] = "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin/lipo"
+    end
+
+    system "autoreconf", "-fvi"
```
dunn: An existing `configure` exists in the distribution; why is this necessary?

webkitgtk24.rb
dunn: Likewise for flex, bison, etc. Basically I'm wondering why this needs so many more deps than the version of webkitgtk in core.

webkitgtk24.rb
dunn: Also geoclue is an X11 formula, so this would need to be

webkitgtk24.rb
eallik: Actually I will remove the `conflicts_with` altogether — it doesn't; I was wrong.

webkitgtk24.rb
eallik: so it's OK that it takes many times longer to build without `env :std` becuase you have to use `deparallelize`? 

Perhaps at least you have a better understanding of why superenv might be causing parallel builds to hang? I'm fine with removing it though and deparallelizing by default — I only wanted it to be fast by default.

webkitgtk24.rb
eallik: `make` didn't seem to work at all without `ENV.deparallelize`, but I can just remove the option then if it's been decided.

webkitgtk24.rb
eallik: Oops.

webkitgtk24.rb
eallik: Well it's a matter of preference really, but I can live with you guys having a different coding convention.

webkitgtk24.rb
eallik: Without geoclue, I believe there will be no geolocation support; that's why I left it on by default.

webkitgtk24.rb
eallik: How can a off-by-default flag cause any problems to those who ignore it?

These flags stand for features that this library _should_ support, but just happen to be broken on OS X — how else do I keep the interface of this formula in line with that of the original package?

webkitgtk24.rb
eallik: Upstream have decided to support them — the stuff's just broken on OS X (but not-broken in the newer 2.10.4 version of the formula).

But OK, if you're policy is to hide features that are (possibly temporarily) broken, I can't really argue — just want the formula to be accepted...

webkitgtk24.rb
eallik: The patch has inline comments if you look closer.

webkitgtk24.rb
eallik: Not sure I understand.

webkitgtk24.rb
eallik: I didn't know I was supposed to not list dependencies that are implied to be put there by Apple.

Is there any documentation where I could read up on all of the feedback you're giving me?

webkitgtk24.rb
eallik: I do not know why this is necessary — that's what the MacPorts port does.

webkitgtk24.rb
dunn: The interface doesn't have to be the same; we aim to have formula build with sensible options by default and only offer options when they're really necessary.

webkitgtk24.rb
dunn: Our documentation could be better :see_no_evil: but here's the (short) section in the Formula Cookbook: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#check-for-dependencies

eallik:
I have an overall question:

If I'm supposed to remove a lot of the comments, all of the broken feature flags, and the reference to the original Portfile, then if somebody at a later point wants to improve the formula, how can they pick up from where I left off?

webkitgtk24.rb
dunn: Later I noticed that one of the inline patches is for `autotools/FindDependencies.m4` to let us use Homebrew ICU, so that's probably why.  Maybe we should let it used the vendored one instead, though. Do you know if it's very out-of-date?

webkitgtk24.rb
eallik: oops, forgot the `@`.

webkitgtk24.rb
DomT4: > How can a off-by-default flag cause any problems to those who ignore it?

Because there will be people who won't ignore it, who then complain it is broken regardless, and then I get to spend 10 hours debugging an option I don't use and didn't want to support and probably still end up cutting the option regardless because it doesn't want to cooperate on OS X.

Sensible defaults save a whole bunch of pain.

webkitgtk24.rb
DomT4: You need to declare it fully as per Alex's syntax. Really would prefer at least giving people an opt-out though by using `=> :recommended`.

webkitgtk24.rb
DomT4: `env :std` in certain cases can have security implications, but more frequently creates a build that differs between the bottles and the from-source build, which then makes life difficult for anyone who hits or has to fix issues.

The vast majority of brew users do not build from source, and consequently will do no compiling.

webkitgtk24.rb
DomT4: i.e. Why are we having to use them? Is there anything upstream saying to use them?

webkitgtk24.rb
DomT4:
```ruby
args = [
  "--prefix=#{prefix}",
  # see https://lists.webkit.org/pipermail/webkit-unassigned/2014-November/645071.html
  "--enable-dependency-tracking",
]
```

DomT4: `%W` just looks neater, but if you have to inline comments, you can still use an `args` block as per.

webkitgtk24.rb
DomT4: I can't realistically ask contributors to read through a 400+ line patch to check each inline comment. It needs to be summarised about the `patch` line. We also need to know if these patches ever made it upstream, MacPorts not being upstream.

webkitgtk24.rb
DomT4: That this formula is going in Versions kinds of implies the chance these problems are ever fixed and released upstream on this branch of releases is close to nil?

DomT4:
If I'm blunt here, if this formula ends up anything like 99% of Versions has, nobody except you will touch this formula again for as long as it exists. That is why we're clamping down [fairly hard](https://github.com/Homebrew/homebrew-versions/commit/992b05c6e55e23a75ee5ae3a684624f7057cc8ad) on what we accept here, and going forwards are removing a bunch of stuff that is insecure and intended to be public facing, hasn't been touched in years, no longer works, etc.

You can include relevant comments, but it can be assumed if you're making major improvements to a formula that 1) You're already partially familiar with it, 2) You know it's a good idea to check out what MacPorts/Fink/Debian/Arch do.

eallik:
OK, fair enough. I guess you're right about somebody else touching this formula being an unlikely event.

eallik:
Just to ping you guys: is there anything still missing? Is the "all checks have failed" a blocker? Or something else?

dunn:
Yeah, it needs to build properly on the CI first.

eallik:
but is the "All checks have failed" somehow related to the formula being built on CI?

webkitgtk24.rb
```diff
@@ -0,0 +1,547 @@
...
+  depends_on "harfbuzz"
+  depends_on "icu4c"
+  depends_on "libxslt"
+  depends_on "libpng"
+  depends_on "libsecret"
+  depends_on "libsoup"
+  depends_on "sqlite"
+  depends_on "webp"
+
+  depends_on "geoclue" => :recommended
```
DomT4: Your current failure is because you're declaring a dependency here but `geoclue` doesn't live in the same tap so `brew` has no idea which `geoclue` you mean. Cross-tap dependencies have to be fully declared.

webkitgtk24.rb
```diff
@@ -0,0 +1,547 @@
...
+    ]
+
+    # https://bugs.webkit.org/show_bug.cgi?id=126329
+    inreplace "Source/JavaScriptCore/API/JSBase.h",
+              /^#define JSC_OBJC_API_ENABLED.*$/,
+              "#define JSC_OBJC_API_ENABLED 0"
+
+    args = %W[
+      --prefix=#{prefix}
+
```
DomT4: Why the whitespacing here?

webkitgtk24.rb
```diff
@@ -0,0 +1,547 @@
...
+      --enable-wayland-target=no
+      --disable-accelerated-compositing
+      --disable-jit
+      --with-svg
+      --without-web-audio
+      --without-webkit2
+      --without-gtk-doc
+    ]
+
+    args.push flag_for("video"),
```
DomT4: Again, would prefer explicit `args` for readability:
```ruby
args << "--with-ducks" if build.with? "ducks"
```
And so on.

webkitgtk24.rb
```diff
@@ -0,0 +1,547 @@
...
+
+    # from the original Portfile
+    ENV.append ["CFLAGS", "CXXFLAGS"], "-ftemplate-depth=256"
+    ENV.append "CXXFLAGS", "-Wno-c++11-extensions"
+    ENV.append "CPPFLAGS", "-DGTEST_USE_OWN_TR1_TUPLE=1"
+
+    # from https://github.com/jralls/gtk-osx-build/blob/master/modulesets/gtk-osx-network.modules#L131-L162
+    ENV.append "CXXFLAGS", "-std=gnu++11"
+
+    # this will make the build much much slower, but if you remove it,
```
DomT4: Don't need this comment. I don't want anything encouraging people to use `:std` env, really.

webkitgtk24.rb
eallik: A-ha, I didn't realize that — it built fine on my machine. WIll fix.

DomT4: Yeah, if you already have the appropriate tap tapped `brew` makes an educated guess.

webkitgtk24.rb
eallik: Quoting you:

> /.../ nobody except you will touch this formula again for as long as it exists

— so I'm a bit confused as to why this review has to be so excruciatingly detail oriented when it comes to what are really matters of preference? I mean, a blank line? Does it really affect anyone in any way? It's there because I wanted to logically separate groups of flags. I can remove the grouping, but is this really important enough to point out on a code file that _"nobody except you will touch /.../ again"_?

webkitgtk24.rb
eallik: Pushed fixes according to this and other comments.

webkitgtk24.rb
DomT4: If I'm building you a house that you're only going to live in for a year and nobody else is going to do any construction work in you'd probably still be a bit miffed if I decided one brick should go in sideways and stick out, just becase. More or less the same principle.

dunn:
OK, here are the build/test failures: http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1107/

mikemcquaid:
Just to save you any more work @eallik: I'm pretty unhappy with this going in given the relatively niche nature and also the number of patches (that will never go upstream) required for this to build.

eallik:
@mikemcquaid but we're talking about homebrew/verisons not the main repo...? given the slow-building nature of it, it would be very useful to have a pre-built bottle available; is there any way for me to get that if I set up my own tap for the formula?

in any case, why is the policy so strict? what harm can happen if this formula is indeed under homebrew/versions? you can always remove it later on if it starts causing any problems...

mikemcquaid:
> @mikemcquaid but we're talking about homebrew/verisons not the main repo...? given the slow-building nature of it, it would be very useful to have pre-built bottles available; is there any way for me to get that if I set up my own tap for the formula?

You could build and publish your own bottle, yeh. See `brew bottle`.

> in any case, why is the stance so strict? what harm can happen if this formula is indeed under homebrew/versions? you can always remove it later on if it starts causing any problems...

We'd rather not add something than add and remove it; the latter is much more painful for users.

eallik:
I guess my view is that the people who need webkitgtk24 know exactly why they need it and what's involved, but then again, if I can set up my own bottles, I guess having an independent tap would work equally well.

I hope at least I won't have to build the bottles on my own hardware and can instead still delegate the building to your existing CI... just possibly without hosting everything in the official repos...? I'd better get to reading the doc though...

mikemcquaid:
I think it'd be best to build this yourself but thanks for the submission and sorry it didn't work out!

2426021684:
@eallik How do you use the formula? I tried 
```
brew install https://raw.githubusercontent.com/eallik/homebrew-versions/8eea95635863d32323771c94740a9f13d2b0c3fb/webkitgtk24.rb
```
and got
```
configure: WARNING: you should use --build, --host, --target
configure: WARNING: you should use --build, --host, --target
configure: WARNING: you should use --build, --host, --target
configure: WARNING: you should use --build, --host, --target
configure: WARNING: unrecognized options: --with-svg, --without-web-audio, --without-webkit2, --without-gtk-doc
checking build system type... Invalid configuration `true': machine `true' not recognized
configure: error: /bin/sh Source/autotools/config.sub true failed
```

k-bx:
@eallik having same issue. Any idea how to resolve this? Thank you!

PR #1081, opened 2015-12-15
libidn2 0.10 (new formula)

wil:
The original `libidn` does not support the latest internationalized domain name (IDN) standards (RFC 5890 - 5893). This package is a compliant implementation.

@DomT4 suggested in Homebrew/homebrew#47005 that this should live here instead of the main repo.


mikemcquaid:

PR #1082, opened 2015-12-15
ack-1.96.rb: add ack 1.96 to homebrew-versions

megahall:
Ack 1.96 is very helpful for people who have existing configurations which are used on large numbers of machines, where Ack 2.x is not always present. In addition Ack 2.x has introduced certain bugs which were not present in Ack 1.96. Therefore nominating this port for inclusion in `homebrew-versions`.

### DIFF
--- a/ack-1.96.rb
+++ b/ack-1.96.rb
@@ -1,0 +1,20 @@
+require 'formula'
+
+class Ack196 < Formula
+  homepage 'http://beyondgrep.com/'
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Passing on this as I don't think it has widespread appeal and we're being a lot stricter with stuff in `homebrew-versions` now. Please consider making your own tap instead. Sorry!

PR #1083, opened 2015-12-15
pgrouting20 2.0.1 (new formula)

sanak:
pgRouting 2.1.0 were released at 2015-09-07 [1], and `pgrouting` formula was updated to the version [2].
But in some cases, pgRouting version 2.0 is necessary to support the applications which are still using old version.

[1] https://lists.osgeo.org/pipermail/pgrouting-users/2015-September/002148.html
[2] https://github.com/Homebrew/homebrew/commit/8e5cfca9be77b262a9a85b3420e44aa733cf34a2

### DIFF
--- a/pgrouting20.rb
+++ b/pgrouting20.rb
@@ -4,6 +4,8 @@ class Pgrouting20 < Formula
   url "https://github.com/pgRouting/pgrouting/archive/pgrouting-2.0.1.tar.gz"
   sha256 "dfb4acddae634e705e680783413317c7482c42c94cd2795dcc359a98356e0e1c"
 
+  conflicts_with "pgrouting", :because => "Differing versions of the same formula"
+
   # work around function name conflict from Postgres
   # https://github.com/pgRouting/pgrouting/issues/274
   patch :DATA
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> But in some cases, pgRouting version 2.0 is necessary to support the applications which are still using old version.

Can you elaborate more on these situations? Thanks!

pgrouting20.rb
```diff
@@ -0,0 +1,85 @@
...
+  desc "PostgreSQL extension to provide spatial routing"
+  homepage "http://www.pgrouting.org"
+  url "https://github.com/pgRouting/pgrouting/archive/pgrouting-2.0.1.tar.gz"
+  sha256 "dfb4acddae634e705e680783413317c7482c42c94cd2795dcc359a98356e0e1c"
+
+  conflicts_with "pgrouting", :because => "Differing versions of the same formula"
+
+  # work around function name conflict from Postgres
+  # https://github.com/pgRouting/pgrouting/issues/274
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4:
Closing as dead. Thank you for your submission to Versions!

PR #1084, opened 2015-12-15
kettle54 5.4.0.1

joshua-rutherford:
This is pretty much a straight lift from the latest 5.X version of the kettle formula in the main repository.  I've updated to that last pre-6.0 version.


mikemcquaid:
Why would you want to use the older version?

joshua-rutherford:
We have developed a number of plugins that break in the newer version so, until we have a chance to update those artifacts, we need to maintain our test environments with our production environments.

joshua-rutherford:
At the same time, I recognize this may not be a wide reaching requirement so if you would prefer we maintain this in our own repo, that will not be an issue either.  Just trying to push upstream if possible.

mikemcquaid:
Thanks @joshua-rutherford. Will let @DomT4 give his thoughts.

joshua-rutherford:
@mikemcquaid, cool.  Since I'll be targeting the latest for my upgrades to our plugins I've opened https://github.com/Homebrew/homebrew/pull/47150 to facilitate that.

DomT4:
No objection to the proposed use case. I think it's valid enough to merit inclusion here, personally.

mikemcquaid:

PR #1091, opened 2015-12-29
mySQL 5.7

aug2uag:


dunn:
This is the version currently in core.  We don't need a Versions formula until there's an upgrade and 5.7 is still needed.

mikemcquaid:
@dunn @DomT4 @xu-cheng: I can see an argument for matching the core version: if you want to make sure that Homebrew upgrading to e.g. 5.8 in your `Brewfile` means you won't be upgraded.

dunn:
But then we have to worry about keep both formulae in sync.

DomT4:
I don't have a problem with this on principle, but my sneaking suspicion is the same as Alex's; people will update the core version and not bother with this one, and one of us will end up doing it.

mikemcquaid:
@DomT4 @dunn Fair points, ignore me. Thanks but closing for the reasons above @aug2uag!

PR #1094, opened 2015-12-30
NiFi 0.4.0 - new formula

apiri:
Creating nifi040 for version 0.4.0 of Apache NiFi (new formula)


mikemcquaid:
Why do you need 0.4.0 rather than 0.4.1?

apiri:
I suppose there is not an explicit need, but was trying to provide formula for all releases generated.  I do know that some folks when they use a certain version within their organization will stick with it and was trying to make it easy for them work against the same should they need to recreate in a development/local environment. 

If this seems outside the scope of homebrew-versions, no problem, can close this issue and will stick to latest of a given minor release.

For future reference, in the case that a prior version has an issue that results in a bugfix/patch version, do folks typically open a PR to purge the replaced version from the repository?

Thanks for any info you can provide!

mikemcquaid:
> I suppose there is not an explicit need, but was trying to provide formula for all releases generated. I do know that some folks when they use a certain version within their organization will stick with it and was trying to make it easy for them work against the same should they need to recreate in a development/local environment.

:cool:. Generally we're being fairly conservative on what we include here and we're sticking to only doing semver major/minor changes (rather than patch releases).

> For future reference, in the case that a prior version has an issue that results in a bugfix/patch version, do folks typically open a PR to purge the replaced version from the repository?

Not totally sure what you mean here, can you elaborate?

apiri:
Thanks for the information. Will close out the associated PR.

My second question is dealing the case that a minor version was both released and merged but then a patched version came out after it had been merged into the repository.  For the sake of example, say 0.3.0 of a formula was merged but the project for which it was made released a patched version, 0.3.1.  In this scenario, should the 0.3.0 version of the formula get removed and the 0.3.1 version added?

mikemcquaid:
> For the sake of example, say 0.3.0 of a formula was merged but the project for which it was made released a patched version, 0.3.1. In this scenario, should the 0.3.0 version of the formula get removed and the 0.3.1 version added?

Gotcha. In that case hopefully we should only have a `foo03` formula and not `foo031` and `foo030`. If a 0.4.0 is released we can leave `foo03` in there. Make sense?

apiri:
Sounds fair. Thanks for the info.

PR #1096, opened 2016-01-01
Ruby 2.2.4

chipp:


DomT4:
Why do you prefer using Homebrew over something like `rbenv` out of curiosity?

bartoszj:
@DomT4 
In my case, sometimes it is useful to install ruby for all users and "override" default version.

ruby22.rb
```diff
@@ -0,0 +1,177 @@
+class Ruby < Formula
+  desc "Powerful, clean, object-oriented scripting language"
+  homepage "https://www.ruby-lang.org/"
+  url "https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.4.tar.bz2"
+  sha256 "31203696adbfdda6f2874a2de31f7c5a1f3bcb6628f4d1a241de21b158cd5c76"
+
+  bottle do
```
DomT4: Please remove this block. We'll add bottles for this tap on pull.

ruby22.rb
```diff
@@ -0,0 +1,177 @@
+class Ruby < Formula
```
DomT4: `class` needs fixing to match the formula name.

ruby22.rb
```diff
@@ -0,0 +1,177 @@
...
+  url "https://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.4.tar.bz2"
+  sha256 "31203696adbfdda6f2874a2de31f7c5a1f3bcb6628f4d1a241de21b158cd5c76"
+
+  bottle do
+    sha256 "0ae4452015623280e369cd6c991f25adf547d9d94fa78c869fa581767069453f" => :el_capitan
+    sha256 "8afc123353ab4373c5cb2b90ad918a92e7416456413f946c46af50ea71935cc2" => :yosemite
+    sha256 "a3efd1af32f998025b70534fc6f505ab082613f7235c36624183d4a7e60221dd" => :mavericks
+  end
+
+  head do
```
DomT4: Please remove the `head` block.

ruby22.rb
```diff
@@ -0,0 +1,177 @@
...
+    url "http://svn.ruby-lang.org/repos/ruby/trunk/"
+    depends_on "autoconf" => :build
+  end
+
+  option :universal
```
DomT4: Please add a `keg_only :provided_by_osx` line one blank line above this line.

DomT4:
Since this is `keg_only` we'll need to mimic the logic of the other Ruby Versions formulae on `rubygems_bindir` as well. Recommend checking out the [Ruby21](https://github.com/Homebrew/homebrew-versions/blob/master/ruby21.rb) formula on that.

ruby22.rb
```diff
@@ -0,0 +1,173 @@
...
+  depends_on "libyaml"
+  depends_on "openssl"
+  depends_on :x11 if build.with? "tcltk"
+
+  fails_with :llvm do
+    build 2326
+  end
+
+  def install
+    system "autoconf" if build.head?
```
UniqMartin: There's no longer a `HEAD` version. Please remove this line.

ruby22.rb
```diff
@@ -0,0 +1,173 @@
...
+    config_file.write rubygems_config
+
+    # Create the sitedir and vendordir that were skipped during install
+    ruby="#{bin}/ruby#{program_suffix}"
+    %w[sitearchdir vendorarchdir].each do |dir|
+      mkdir_p `#{ruby} -rrbconfig -e 'print RbConfig::CONFIG["#{dir}"]'`
+    end
+
+    # Create the version-specific bindir used by rubygems
+    mkdir_p "#{rubygems_bindir}"
```
UniqMartin: This is the line that failed on the build bot because `rubygems_bindir` is not defined in the formula. Please add [these lines from `ruby21`](https://github.com/Homebrew/homebrew-versions/blob/faf44a69f101ec9bedef2a7ad6799c7b2f7db1cf/ruby21.rb#L111-L113) below the `abi_version` method to fix that. Thanks!

ruby22.rb
```diff
@@ -0,0 +1,173 @@
...
+      def self.default_path
+        if Gem.user_home && File.exist?(Gem.user_home)
+          [user_dir, default_dir, private_dir]
+        else
+          [default_dir, private_dir]
+        end
+      end
+
+      def self.default_bindir
+        "#{HOMEBREW_PREFIX}/bin"
```
UniqMartin: Please use `rubygems_bindir` here instead of this path. See [the `ruby21` formula](https://github.com/Homebrew/homebrew-versions/blob/faf44a69f101ec9bedef2a7ad6799c7b2f7db1cf/ruby21.rb#L169-L171) for how this should look like. Thanks!

ruby22.rb
```diff
@@ -0,0 +1,173 @@
...
+      end
+
+      def self.ruby
+        "#{opt_bin}/ruby#{"22" if build.with? "suffix"}"
+      end
+    end
+    EOS
+  end
+
+  test do
```
UniqMartin: Please add a `caveats` method above the `test do` block. Once again, you can [copy it from `ruby21`](https://github.com/Homebrew/homebrew-versions/blob/faf44a69f101ec9bedef2a7ad6799c7b2f7db1cf/ruby21.rb#L180-L186). This will tell users where Gems installed with this version of Ruby will end up, as they can't go into `HOMEBREW_PREFIX/bin` because that is already occupied by the `ruby` formula in core. Thanks!

ruby22.rb
UniqMartin: Could you please address this code comment, too? (Also, it would help us if you could acknowledge our feedback and tell us when you've incorporated it and pushed a new commit, as we usually don't get notified if you just push a new commit. Thanks!)

chipp: Fixed. Sorry for silence from my side. :)

UniqMartin:
Merged in 7da054230ce063f6145f6a2818710b18ac8b5079. Thank you for this contribution to Homebrew, @chipp! :tada:

PR #1097, opened 2016-01-04
ruby187-brew 1.8.7 (new formula)

DomT4:
I'm submitting this as a new formula because the modifications required to standard Ruby 1.8.7 to incorporate the vendored OpenSSL are kind of heavy and crappy to push onto users including:

* Making everyone who can't pour bottles build OpenSSL and then Ruby from source.
* Making the formula unavailable to anyone using Mountain Lion or below thanks to OpenSSL's new and as-yet unfixed Clang detection bug. *(Technically only to Lion and Mountain Lion users I think, but we don't have specific depends_on syntax to exclude only two OS X versions AFAIK)*.
* Extra build dependencies for OpenSSL.
* Mandating they keep a vendored OpenSSL on their system, and relying on people to update quickly if something horrible happens with the current OpenSSL's security.

Whether or not this is the right tap to place something like this is a different question. We may prefer to entertain putting this in a `homebrew-dev` tap or something and bottling it there.

This formula is needed because due to Mavericks CI using `ruby187` from this tap every time we push an OpenSSL change/update/etc through CI it removes the OpenSSL install (obviously) and then `ruby187` which is built against it panics wildly that it can no longer find that OpenSSL and breaks the build, badly, for every formula, like this...
```
==> brew fetch --retry ooniprobe
==> FAILED
Error: dlopen(/usr/local/Cellar/ruby187/1.8.7-p374_1/lib/ruby/1.8/i686-darwin13.4.0/digest/md5.bundle, 9): Library not loaded: /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib
  Referenced from: /usr/local/Cellar/ruby187/1.8.7-p374_1/lib/ruby/1.8/i686-darwin13.4.0/digest/md5.bundle
  Reason: image not found - /usr/local/Cellar/ruby187/1.8.7-p374_1/lib/ruby/1.8/i686-darwin13.4.0/digest/md5.bundle
Please report this bug:
    https://git.io/brew-troubleshooting
/usr/local/Cellar/ruby187/1.8.7-p374_1/lib/ruby/1.8/i686-darwin13.4.0/digest/md5.bundle
```
Which means we can't pass *any* OpenSSL CI builds on Mavericks.

If we particularly want to I'll write this into the `ruby187` formula, but it feels a heavy change requiring a revision which would break existing installs on certain platforms and so on, so forth.


DomT4:
> == /usr/local/Library/Taps/homebrew/homebrew-versions/ruby187-brew.rb ==
W: 53:  7: Method definitions must not be nested. Use lambda instead.
W: 60:  7: Method definitions must not be nested. Use lambda instead.

Tis the only failure. I kept it that way because it matches the formula in the core, and because I am horrible at lambdas.

UniqMartin:
This formula is a great start to alleviate many of the issues we have with running `ruby187` on our CI!

> Whether or not this is the right tap to place something like this is a different question. We may prefer to entertain putting this in a `homebrew-dev` tap or something and bottling it there.

Yes, we can decide on the placement later. I'm very much in favor of putting it in a separate tap to get across that this is for internal use only, hence `homebrew-internal` and `homebrew-ci` also come to mind.

mikemcquaid:
Would any of this be solved by removing `ruby187` from the `$PATH`? I definitely think this should go in another tap, probably `homebrew-dev`.

UniqMartin:
> Would any of this be solved by removing `ruby187` from the `$PATH`?

No, that's a separate issue (mostly with Ruby Gems) that also needs to be ironed out. I once started looking at what would need to change to make it work without adding it to `$PATH`, but got distracted …

mikemcquaid:
> No, that's a separate issue (mostly with Ruby Gems) that also needs to be ironed out. I once started looking at what would need to change to make it work without adding it to $PATH, but got distracted …

It seems in that case we could perhaps add it to the PATH just for the commands that require the RubyGems stuff (e.g. `audit`, `style`, `tests`).

DomT4:
Aye, as Martin says. This PR primarily resolves the problem we've been having with every OpenSSL PR since we adopted 1.8.x on Mavericks Jenkins, which is the unlinking of OpenSSL causing this Ruby to break. I suspect the same thing might happen with `readline` but since that hasn't been updated or revisioned since September 2014 it's not as pressing as OpenSSL, which gets done regularly.

mikemcquaid:
> I suspect the same thing might happen with readline but since that hasn't been updated or revisioned since September 2014 it's not as pressing as OpenSSL, which gets done regularly.

Could you try it and see? If so, perhaps vendor `readline` too?

DomT4:
Sure. I'll file a PR in the core to fake revision `readline` and see if Mavericks panics similarly.

mikemcquaid:
@DomT4 :+1:

UniqMartin:
> Sure. I'll file a PR in the core to fake revision `readline` and see if Mavericks panics similarly.

I'd say this is unlikely, as `libreadline.6.dylib` is only loaded after a `require "readline"` and that is probably not happening in our use case (most uses of `readline` in a stock Ruby are in `irb`).

UniqMartin:
> It seems in that case we could perhaps add it to the PATH just for the commands that require the RubyGems stuff (e.g. `audit`, `style`, `tests`).

Yes, that would be a workaround until a better solution is found. We could make that a part of [`Utils.install_gem_setup_path!`](https://github.com/Homebrew/homebrew/blob/d947a9dea238df4fdf460ba28e856725342090e4/Library/Homebrew/utils.rb#L232-L253) which is tweaking `PATH` for the Gems anyway, i.e., make it additionally add the path of the currently running Ruby binary.

DomT4:
> I'd say this is unlikely, as libreadline.6.dylib is only loaded after a require "readline"

That'd be helpful if that proved the case. In that situation we would need to revision this Ruby against an updated API-incompatible or library-version-shifted (if a change that drastic happens, at this point) new `readline` release presumably, but we could avoid the vendor at least.

DomT4:
>  I definitely think this should go in another tap, probably homebrew-dev.

https://github.com/Homebrew/homebrew-dev-tools or a new tap called `homebrew-dev`?

UniqMartin:
> https://github.com/Homebrew/homebrew-dev-tools or a new tap called `homebrew-dev`?

I'm in favor of the latter and would prefer if the existing tap continued to focus on external commands targeted at maintainers.

mikemcquaid:
> https://github.com/Homebrew/homebrew-dev-tools

Yup, that one :+1:

DomT4:
`readline` PR passed (well, it failed, but for unrelated reasons) so we're alright there. Phew. I'll roll this over to `dev-tools`. Do I need to hook it up to Jenkins or anything so we can bottle this?

DomT4:
(Or more accurately do you need to hook it up to Jenkins? :wink:. Not entirely sure how to hook a new project up to the CI here).

mikemcquaid:
@DomT4 I'll take a look.

DomT4:
Cheers!

mikemcquaid:
Cross-posting from https://github.com/Homebrew/homebrew/issues/47694: another solution to this problem is to install Ruby 1.8.7 in a non-`/usr/local` prefix on both Jenkins and Travis CI, possibly just installing it with e.g. `ruby-build`.

UniqMartin:
To pick up on Mike's cross-posting: I also had a few thoughts in https://github.com/Homebrew/homebrew/pull/47694#issuecomment-169196535.

mikemcquaid:
The Jenkins bot now uses a local install: http://bot.brew.sh/job/Homebrew%20Testing/916/version=mavericks/testReport/brew-test-bot/mavericks/config/. Probably the best solution for the Travis bot is to build the various bottles and upload them somewhere but at least we're unblocked with Jenkins for now.

DomT4:
Thanks, I was still thinking about how best to handle that.

I can build bottles manually for a non-default prefix and then add a step to our Travis config that creates a new `brew` install in that location and pours the bottle from our Bintray Mirror repo or something.

Alternatively I can build from source locally with a vendored OpenSSL & readline, wrap that in a tarball, upload the tarball to Bintray and then we just fetch & untar it in that location. Essentially like a bottle, except without needing Homebrew there to handle the install step.

The former is probably more easily consistent, but happy to stick both options on the table.

mikemcquaid:
I'd consider not using a bottle at all and using `ruby-build` to built it initially and then install it into the same prefix on Travis.

mikemcquaid:
(as you can coax `ruby-build` into installing e.g. `openssl`, `libyaml` etc. too into the same prefix)

UniqMartin:
> The Jenkins bot now uses a local install:

:+1:

DomT4:
> I'd consider not using a bottle at all and using ruby-build to built it initially and then install it into the same prefix on Travis.

Fair enough. Give me a bit and I'll throw something together.

DomT4:
Closing here because I'm about to propose an alternative elsewhere.

DomT4:
For historical clarity: https://github.com/Homebrew/homebrew-dev-tools/pull/3

PR #1099, opened 2016-01-04
gcc6 6-20160103 (new formula)

manphiz:
Add gcc6 as now that GCC trunk have reached stage 3.


gcc6.rb
```diff
@@ -0,0 +1,209 @@
...
+  test do
+    (testpath/"hello-c.c").write <<-EOS.undent
+      #include <stdio.h>
+      int main()
+      {
+        puts("Hello, world!");
+        return 0;
+      }
+    EOS
+    system bin/"gcc-5", "-o", "hello-c", "hello-c.c"
```
mikemcquaid: `gcc-5` needs updated.

gcc6.rb
```diff
@@ -0,0 +1,209 @@
...
+  def pour_bottle?
+    MacOS::CLT.installed?
+  end
+
+  # GCC bootstraps itself, so it is OK to have an incompatible C++ stdlib
+  cxxstdlib_check :skip
+
+  # Fix for libgccjit.so linkage on Darwin.
+  # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64089
+  patch :DATA
```
mikemcquaid: Looks like this was merged upstream? Would be good to use (or at least link to) the upstream commit instead.

gcc6.rb
manphiz: Indeed, will fix.

gcc6.rb
manphiz: I tried removing the patch but it still fails as ld on darwin doesn't recognize `--version-script` option. So I guess we have to keep this patch a little longer. Will keep tracking this bug.

mikemcquaid: Is there an upstream commit we can backport and/or point to instead?

DomT4: I missed Mike's comment as it was hidden under an `outdated diff` message. Feel free to file a follow-up PR if there's something we can do there.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/bf0e0a0c97218e5703fa4cd7984471ca46d2680a. Thank you @manphiz!

gcc6.rb
manphiz: -----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256



On 1/5/16 12:19 AM, Mike McQuaid wrote:
> In gcc6.rb 
> <https://github.com/Homebrew/homebrew-versions/pull/1099#discussion_r48820946>:
>
> 
>> +  end + +  fails_with :llvm + +  # The bottles are built on
>> systems with the CLT installed, and do not work +  # out of the
>> box on Xcode-only systems due to an incorrect sysroot. +  def
>> pour_bottle? +    MacOS::CLT.installed? +  end + +  # GCC
>> bootstraps itself, so it is OK to have an incompatible C++
>> stdlib +  cxxstdlib_check :skip + +  # Fix for libgccjit.so
>> linkage on Darwin. +  #
>> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64089 +  patch
>> :DATA
> 
> Is there an upstream commit we can backport and/or point to
> instead?
> 

Last time I checked the GCC bugzilla, upstream is yet to reach a
consensus on this issue. Will keep following on this.

> — Reply to this email directly or view it on GitHub 
> <https://github.com/Homebrew/homebrew-versions/pull/1099/files#r48820946>.
>
-----BEGIN PGP SIGNATURE-----
Version: GnuPG/MacGPG2 v2

iQIcBAEBCAAGBQJWjP+IAAoJEC3pZe1jglyTHskP/1as5seT3jiL24Ri/jjdOtcE
CA+N/vcSNQ4j/WPTHC59TCDA0WmxvEywQkB/NNIKb7P1LOhv7qbFAXOsIAjUZT9t
PYIusvQoJMpro/yZdcThokelTJVBGUunaHuyW6AslHy39b/pKwRGVMrcZZHACTj+
J+0qYyU52Pgn5QgcFJGQIExiFTJT4WQ4dG6EjmIPBWRynXGc7n2Ht+XfjPmy7XCT
XiSnzKkPaMUBhYxbTqXLOzlWWpS10VQSP1ZkgGbsoVuAkX3S6wqGg4yth452XkvM
6ldvi7uwRV3c16ehxF+uhkK52f/ox75APVdzDE8hflWadBNvUrQQjvF0LYAPWZTy
TkDrFpfPx918e9I3lkL3AvEfcKbN5hJdKSbhA1UmzR9bcKmprW+ULzJ2QFD/hPtV
ukPt/g35CBE/qri/xBJ3A1qJijwdgu/vtO26hRROpya2CAdJFOmfH45miR9d+Mez
2YfwjlX8ZYlZVUrrYIM9OLpDJPlSQypHCyWs+hNvBUrx1F0QqWjKGQqTN7zY9mRq
3mFQUl1PAobrG5UHxqtSnEnVT28AdbSQi71qs2vouD+Aun6VRZK2HE9DMvP0C8Ie
iHljK4ISsbQHgq8guVB99YJWb8dt+8Hb2IMq8NgXiwtyEeaSbd35dTjcWwiAmo0m
q8Ed5sFiS9DDJnjruyWa
=Wtcq
-----END PGP SIGNATURE-----

PR #1102, opened 2016-01-06
Add selenium-server-standalone 2.43.1 version

tilthouse:
I needed this particular old version of selenium-server-standalone, so here is a formula for it.


mikemcquaid:
Can you elaborate why the newer version doesn't work for you? Just trying to get a feeling for use-cases of things in versions. Thanks!

Issue #1103, opened 2016-01-09
postgresql94


joelmichael:
I think lots of people will probably be wanting this now that 9.5 is out and on the main homebrew. Can someone make it

Maybe I can be the change I want to see in the world but I'm not sure I'm ready and someone probably knows how to do this easily

mikemcquaid:
Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

njamaleddine:
Hey, I closed out my PR, and respectfully decided to leave it to the OP after reading the above comment. Keep me updated on the progress :smile:

richardkmichael:
Somehow I accidently updated my Postgres 9.4.5 -- I thought I had it pinned but perhaps I did something wrong (most frequent explanation!) or homebrew ignored my pin.

In any case, I wound up with `9.5` and needed 9.4.5 back immediately, but didn't see it available anywhere.  I'm not sure how homebrew handles previous versions -- it appears previous-version formulae are eventually moved here?  Again, perhaps I've misunderstood homebrew: admittedly, I've paid scant attention to the details, as Homebrew usually does what I want.

To get 9.4.5 back, I've naively done:

```
brew uninstall postgresql # Remove the 9.5 which was upgraded.
brew clean --force -s
cd /usr/local
git checkout -b revert-postgres
git show f8509e62904a055f085579aed47fca1faa7a810f:Library/Formula/postgresql.rb > postgresql.rb
brew install postgresql
```

It compiled and installed without problems.

The OP of this PR seems to have opted out.  Could/should I simply PR Homebrew/homebrew@f8509e62904a055f085579aed47fca1faa7a810f, with a few changes; i.e., `class Postgresql94 < Formula`?  Also, I'm not sure what to do about the bottle `sha256` values.

DomT4:
Closing in favour of https://github.com/Homebrew/homebrew-versions/pull/1105.

PR #1106, opened 2016-01-11
Add Postgresql94 at 9.4.5

joelmichael:
Does what it says on the tin


joelmichael:
Ok I made a commit #1103

I just pulled the latest 9.4.5 formula from the main homebrew repo and changed the version name and removed the devel block. The tests are taking forever to run (I guess there is a lot of competition) but in theory this should work

joelmichael:
Well I gave it a shot but it doesn't work and I don't know why, and it takes 2 hours to run the tests

PR #1107, opened 2016-01-12
gcc-6-20160110

roederw:


gcc6.rb
```diff
@@ -21,15 +21,9 @@ def osmajor
 
   desc "The GNU Compiler Collection"
   homepage "https://gcc.gnu.org"
-  url "ftp://gcc.gnu.org/pub/gcc/snapshots/6-20160103/gcc-6-20160103.tar.bz2"
-  sha256 "1a5cd4a08b8579ebbab7667c1ebc01a51958a64cf1898720db28c6b8cd69155e"
-
-  bottle do
```
dunn: Leave the bottle block, please! Our CI bot handles updating it.

dunn:
Also:
```
        Error: 1 problem in 1 formula
homebrew/versions/gcc6:
 * 26: Trailing whitespace was found

roederw:
Leave the outdated sha values as well?

jenshnielsen:
Is there any reason why Fortran is disabled per default unlike the main gcc5 formula?

DomT4:
> Leave the outdated sha values as well?

Yeah. We update them automatically with the new ones on merge.

> Is there any reason why Fortran is disabled per default unlike the main gcc5 formula?

I *suspect* by default the `gfortran` binary isn't versioned and would conflict with the main `gcc` formula. Don't quote me on that though. To my knowledge we've never built the `fortran` option by default in Versions.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/77023d82f0a8a600f3b6ba67f2783bc212f3f2c7. Thank you for your contribution to Homebrew @roederw; we appreciate it! :smiley_cat:

mikemcquaid:
> I suspect by default the gfortran binary isn't versioned and would conflict with the main gcc formula. Don't quote me on that though. To my knowledge we've never built the fortran option by default in Versions.

I think this is probably just a legacy thing as we didn't use to build it in Homebrew core either. :+1: to PR to build `gfortran` by default.

jenshnielsen:
@mikemcquaid Thanks. Im happy to put one together and see how it works out

DomT4:
I've no opposition if there's not actually a reason we don't build it here. If anything Versions formulae *tend* to be more fully-featured than the core, for example the `llvm` formulae here include Clang by default whereas the core does not.

PR #1108, opened 2016-01-12
GCC6: Build gfortran per default

jenshnielsen:
Matches unversioned gcc formula in core. As discussed in #1107
Looking at the formula in core https://github.com/Homebrew/homebrew/blob/master/Library/Formula/gcc.rb#L161 the non versions executable is created as a symlink to the versioned one so I expect that this should work without versioning issues.


DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/a509daac1ae833475b956e756c1b05a4255d5e3c. Thank you for your contribution to Homebrew @jenshnielsen; we appreciate it! :smiley_cat:.

Didn't bother fixing it here, but formula names in the commit message should be all lowercase. It's terribly minor, but the way we like things, heh.

PR #1110, opened 2016-01-13
gcc5 - default gfortran update

roederw:
Moving to use the most recently gcc-5 snapshot, since gcc 5.3 is now in core. Also, adding Fortran changes to follow a recent merge to gcc-6 as discussed in #1108.


DomT4:
We prefer to keep these formulae stable; they aren't designed to stay on the bleeding edge (people should use `head` from the core for that). The compilers in Versions tend to be used by people who need stability and would have issues if/when Homebrew core upgrades to a future incompatible GCC.

I'm happy to take the fortran change, though.

roederw:
I can respect that. But, shouldn't we have a gcc53 that is up to date and a gcc52 for stability?

DomT4:
We try to only track stable releases from each major branch or it gets a bit much trying to keep on top of everything, which Versions already suffers from really quite badly sadly.

People are free to support whatever they wish to in their own tap, but Versions isn't really intended to be bleeding edge in the majority of cases.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/9e6bad227d7ef147e0b41cd530133a75d7e489f7. Thank you for your contribution to Homebrew @roederw; we appreciate it! :smiley_cat:

PR #1114, opened 2016-01-14
new formula: flex 2.5.39

houtanb:
Add formula for old version of `flex`. Needed for the compilation of `dynare.rb` in `homebrew-science`.

### DIFF
--- a/flex25.rb
+++ b/flex25.rb
@@ -1,0 +1,45 @@
+class Flex < Formula
+  desc "Fast Lexical Analyzer, generates Scanners (tokenizers)"
+  homepage "http://flex.sourceforge.net"
+  url "https://downloads.sourceforge.net/flex/flex-2.5.39.tar.bz2"
+  sha256 "add2b55f3bc38cb512b48fad7d72f43b11ef244487ff25fc00aabec1e32b617f"
+
+  bottle do
+    sha256 "31607eeace06e7460d03b2a0bfb9b13f3ca5b0407eb3e80d75d3b4f1ec43ba0f" => :el_capitan
+    sha1 "a1f8a76465942c6187b00dfa931ac56f22b7ba00" => :yosemite
+    sha1 "533d4351154274d9d68f517b85e1de8e524b2bd5" => :mavericks
+    sha1 "524d471123dbd4ca812e3fc3778afd865df616e6" => :mountain_lion
+  end
+
+  keg_only :provided_by_osx, "Some formulae require a newer version of flex."
+
+  depends_on "gettext"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-shared",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.flex").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


flex25.rb
```diff
@@ -0,0 +1,45 @@
+class Flex < Formula
```
mikemcquaid: Needs to be called `Flex25`

flex25.rb
```diff
@@ -0,0 +1,45 @@
...
+  homepage "http://flex.sourceforge.net"
+  url "https://downloads.sourceforge.net/flex/flex-2.5.39.tar.bz2"
+  sha256 "add2b55f3bc38cb512b48fad7d72f43b11ef244487ff25fc00aabec1e32b617f"
+
+  bottle do
+    sha256 "31607eeace06e7460d03b2a0bfb9b13f3ca5b0407eb3e80d75d3b4f1ec43ba0f" => :el_capitan
+    sha1 "a1f8a76465942c6187b00dfa931ac56f22b7ba00" => :yosemite
+    sha1 "533d4351154274d9d68f517b85e1de8e524b2bd5" => :mavericks
+    sha1 "524d471123dbd4ca812e3fc3778afd865df616e6" => :mountain_lion
+  end
```
mikemcquaid: Need to remove the `bottle do` block.

mikemcquaid:
@houtanb It's usually preferable to vendor formulae as `resource`s if they are needed by a single formula.

houtanb:
@mikemcquaid, I'm a bit lost, which is surely due to my lack of ruby knowledge. I have tried to install flex 2.5 as a resource in [`dynare.rb`](https://github.com/Homebrew/homebrew-science/blob/master/dynare.rb), to no avail. I modify the `head do` block to be:
```
  head do
    url "https://github.com/DynareTeam/dynare.git"
    depends_on "automake" => :build
    depends_on "bison"    => :build
    depends_on "gettext"  => :build

    resource "flex" do
      url "https://downloads.sourceforge.net/flex/flex-2.5.39.tar.bz2"
      sha256 "add2b55f3bc38cb512b48fad7d72f43b11ef244487ff25fc00aabec1e32b617f"
    end
  end
```
which I think is correct. However, I am having trouble in the `def install` block. Here's what I have:
```
  def install
    args = %W[
      --disable-debug
      --disable-dependency-tracking
      --disable-silent-rules
      --prefix=#{prefix}
    ]
    head do
      resource("flex").stage do
        system "./configure", "--disable-dependency-tracking",
               "--disable-shared",
               "--prefix=#{prefix}"
        system "make", "install"
      end
      args << "LEX=#{prefix}/bin/flex"
    end
```
What I need to do, it seems, is to install flex 2.5 to the `buildpath`, then include the path to the flex 2.5 binary in the `LEX` variable so that `dynare`'s `./configure` can find it.

mikemcquaid:
@houtanb In the `install` block you need to use `if build.head?` instead of `head do`

DomT4:
:+1: for vendoring where necessary.

houtanb:
Closing as I updated `dynare.rb` to use `flex` as a resource. Pull request in Homebrew/homebrew-science#3414

Thanks for the help @mikemcquaid

PR #1115, opened 2016-01-14
apache-spark152 1.5.2 (new formula)

heuermh:


apache-spark152.rb
```diff
@@ -0,0 +1,24 @@
+class ApacheSpark141 < Formula
```
mikemcquaid: This name should be `ApacheSpark152`

heuermh: thanks!  rebased with fix

heuermh:
Closing in favor of #1157, thanks @jimhe!

PR #1119, opened 2016-01-18
python2.7.10

xcolwell:
The python 2.7.11 in brew introduces this error into some programs:

```
ImportError: dlopen(/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload/_io.so, 2): Symbol not found: __PyCodecInfo_GetIncrementalDecoder
```

2.7.10 does not have this error.

### DIFF
--- a/python2.7.10.rb
+++ b/python2.7.10.rb
@@ -1,0 +1,393 @@
+# from cff239addef86cc2bdeac6d97bd57992540f2ace python.rb
+class Python < Formula
+  desc "Interpreted, interactive, object-oriented programming language"
+  homepage "https://www.python.org"
+  head "https://hg.python.org/cpython", :using => :hg, :branch => "2.7"
+  url "https://www.python.org/ftp/python/2.7.10/Python-2.7.10.tgz"
+  sha256 "eda8ce6eec03e74991abb5384170e7c65fcd7522e409b8e83d7e6372add0f12a"
+  revision 2
+
+  bottle do
+    revision 3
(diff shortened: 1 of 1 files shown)


mikemcquaid:
We don't want to accept differing patch versions in homebrew-versions, I'm afraid.

xcolwell:
Just curious - why?

By the way, should this also work?

```
brew tap xcolwell/homebrew-versions
brew install xcolwell/homebrew-versions/python2.7.10
```

But really `xcolwell/homebrew-versions` gets written as `xcolwell/versions` and no matter which I use, I get no install and the same output:

```
Error: No available formula with the name "xcolwell/versions/python2.7.10"
==> Searching for similarly named formulae...
This similarly named formula was found:
xcolwell/versions/python2.7.10
To install it, run:
  brew install xcolwell/versions/python2.7.10
==> Searching taps...
Error: No formulae found in taps.
```

mikemcquaid:
@xcolwell Hopefully this explains a bit: https://github.com/Homebrew/homebrew-versions#acceptable-formulae

DomT4:
See also:

* https://github.com/Homebrew/homebrew-versions/pull/513#issuecomment-75676199
* https://github.com/Homebrew/homebrew-versions/pull/967#issuecomment-140919714
* https://github.com/Homebrew/homebrew-versions/pull/882#issuecomment-119012686

PR #1120, opened 2016-01-19
gcc-6-20160117

roederw:
Updated to the most recent build of gcc6. Updating every revision since they build towards a more stable build, in theory.


roederw:
I discovered an issue and will re-open when I find a fix.

gcc6.rb
```diff
@@ -21,8 +21,8 @@ def osmajor
 
   desc "The GNU Compiler Collection"
   homepage "https://gcc.gnu.org"
-  url "ftp://gcc.gnu.org/pub/gcc/snapshots/6-20160110/gcc-6-20160110.tar.bz2"
-  sha256 "943ed6c5ecb2fcd1c98e450815fb235a4dc00dcd0299ae985f41e14bdaf0653b"
+  url "ftp://gcc.gnu.org/pub/gcc/snapshots/6-20160117/gcc-6-20160117.tar.bz2"
+  sha256 "3c087e6f015a93e1bf2ab9986c3cd389de33442c6766a6cc641be36534c31c8a"
   revision 1
```
dunn: For future reference you want to remove the `revision` line when updating the version.

roederw: It is automagically added by BrewBot once the pull request is accepted?

dunn: No, but we only need a `revision` line if we need to force an upgrade without actually bumping the version.

roederw: Thanks for the information :)

PR #1122, opened 2016-01-20
Create file opus-tools-0.1.6.rb

kickermeister:
The version 0.1.6 is the last that supports the "--uncoupled" flag for libopus. Works fine with current version of libopus 1.1.2


dunn:
A number of audit failures (FYI you can check these locally with `brew audit --strict --online <formula>`):
```
== /usr/local/Library/Taps/homebrew/homebrew-versions/opus-tools.rb ==
C:  1:  9: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  4: 12: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  5:  7: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  6:  8: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C:  8:  8: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C: 15: 14: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C: 16: 14: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
C: 17: 14: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

1 file inspected, 8 offenses detected
Error: 8 problems in 1 formula

==> audit problems
homebrew/versions/opus-tools:
 * Formula name conflicts with existing core formula.
 * `require 'formula'` is now unnecessary
 * :automake is deprecated. Usage should be "automake"
 * :libtool is deprecated. Usage should be "libtool"
 * Use `system "make", "install"` instead of `system "make install"`

File: opus-tools.rb [hunk 1]
```diff
@@ -0,0 +1,25 @@
+require 'formula'
+
+class OpusTools < Formula
```
dunn: This would be `class OpusTools01` if this is a formula for the 0.1.x branch, or `class OpusTools016` if it's going to be only this version.

apjanke:
Closing due to inactivity.

PR #1126, opened 2016-01-21
perconaserver55: fix libexec install

anfleene:
The percona-server55 formula is failing due to a missing libexec directory. This fixes it.

### DIFF
--- a/percona-server55.rb
+++ b/percona-server55.rb
@@ -117,6 +117,7 @@ class PerconaServer55 < Formula
     ln_s "#{prefix}/support-files/mysql.server", bin
 
     # Move mysqlaccess to libexec
+    libexec.mkpath
     mv "#{bin}/mysqlaccess", libexec
     mv "#{bin}/mysqlaccess.conf", libexec
   end


percona-server55.rb
```diff
@@ -109,6 +109,7 @@ def install
     ln_s "#{prefix}/support-files/mysql.server", bin
 
     # Move mysqlaccess to libexec
+    libexec.install Dir["*"]
```
DomT4: Don't we want to `libexec.mkpath` here rather than dumping whatever is left in the `buildpath` at this point into `libexec`?

anfleene: Ah yeah that's right. Sorry.

anfleene: Fixed.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/ce4559f40f6b15fe4278d0668f75df124fa4af44. Thank you for your contribution to Homebrew @anfleene! :smiley_cat:

PR #1129, opened 2016-01-22
Added cassandra 2.2.3

rstjohn:


DomT4:
We're trying to clamp down on point release formulae. Why do you need `2.2.3` and can't use `2.2.4`?

rstjohn:
Version 2.2.4 did not work for us.  Everyone that brew installed cassandra22 where unable to use the SELECT statement from the cqlsh.  The SELECT statement always returned with an error:

'PageResult' has no attribute 'has_more_pages' 

Neither version 2.2.3 nor 2.2.2 had the same issue.  As such, we cannot brew install 2.2.4.

ckhall:
@rstjohn I submmited PR #1131 yesterday, which fixes the problem with 2.2.4.  @mikemcquaid accepted it yesterday. you should be good to go.

DomT4:
Closing per the above, but do let us know if you have further trouble. Thanks!

PR #1134, opened 2016-01-26
Pandoc cabal_clean_lib fixup

mikemcquaid:
This method was removed from the DSL and these formulae still used it.

CC @DomT4 @tschoonj


DomT4:
:+1:. Could do with bottles for El Cap here as well when pulled.

mikemcquaid:
@DomT4 From local testing: I suspect they won't build.

mikemcquaid:
https://github.com/Homebrew/homebrew/pull/48479 is another fix for this but both would be best.

mikemcquaid:
As expected: these fail to build so I'm just going to remove them. We should probably avoid adding old versions of stuff that uses `cabal` as they break so easily.

PR #1139, opened 2016-01-27
fixed mpfr2 to run 'make check' correctly when in linux. 

hradec:
make check was failing in linux since gmp4 shared libs weren't in LD_LIBRARY_PATH.


mikemcquaid:
This repository is only for Homebrew and not Linuxbrew.

PR #1158, opened 2016-02-10
dummy change astyle204: upgrade stable

apjanke:
DO NOT MERGE. This is just a dummy change for use in
testing homebrew/homebrew#48472.


DomT4:
Presuming safe to close via https://github.com/Homebrew/homebrew/pull/48580#issuecomment-183548767 but shout at me otherwise!

PR #1159, opened 2016-02-10
fix lua53 pkg-config library name

Habbie:


DomT4:
I'm torn on whether we should do this or just create a `liblua5.3.dylib`. Lua does the former, Lua51 does the latter.

Habbie:
Oh, interesting. I'll try to have a closer look tomorrow.

Habbie:
I see the difference, indeed. Any idea why the lua51 formula has that symlink, instead of fixing the `.pc`? Non pkg-config users expecting it?

DomT4:
Ah, it may well be [our fault](https://github.com/Homebrew/homebrew/blob/be05c000494493a8193a543dc5d78402c8b415d2/Library/Formula/lua51.rb#L163). That patch has been around forever, but we apparently wrote it slightly differently for `lua51` and `lua53`. I'm fairly happy to accept this then. We can re-review if anyone complains, I guess.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/028e86632561f01d4b9451ca6f7e77f25dff671a. Thanks @Habbie! :beer:

PR #1164, opened 2016-02-13
elasticsearch14 1.4.4 (new formula)

artursouza:


DomT4:
This was removed previously in https://github.com/Homebrew/homebrew-versions/commit/abc8851b4c0bb7b92c82764b854d147687ad23ff by one of the upstream devs because they no longer support it. I'm not keen on re-adding it given that.

Given this doesn't need bottling at all though, it'd be really easy and quick to support in your own tap, which you can find documentation on [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) if desired. Thanks for the submission!

artursouza:
Sure. Thanks.

PR #1171, opened 2016-02-24
coq 8.4pl6

ilovezfs:
Coq 8.4pl6 was the last patch release of Coq 8.4 in Homebrew master
before the upgrade to 8.5 and the last patch release of the 8.4 series
released by upstream: homebrew/homebrew@11fb166

The immediate purpose for having Coq 8.4 in homebrew/versions is to
support the CompCert formula in Homebrew master. CompCert has not yet
been updated to support Coq 8.5, though a third party fork does exist
offering such support:

https://github.com/jhjourdan/CompCert/tree/coq8.5-compcert2.6

### DIFF
--- a/coq84.rb
+++ b/coq84.rb
@@ -1,0 +1,51 @@
+class Camlp5TransitionalModeRequirement < Requirement
+  fatal true
+
+  satisfy(:build_env => false) { !Tab.for_name("camlp5").with?("strict") }
+
+  def message; <<-EOS.undent
+    camlp5 must be compiled in transitional mode (instead of --strict mode):
+      brew install camlp5
+    EOS
+  end
+end
+
+class Coq84 < Formula
+  desc "Proof assistant for higher-order logic"
+  homepage "https://coq.inria.fr/"
+  url "https://coq.inria.fr/distrib/V8.4pl6/files/coq-8.4pl6.tar.gz"
+  version "8.4pl6"
+  sha256 "a540a231a9970a49353ca039f3544616ff86a208966ab1c593779ae13c91ebd6"
+
+  bottle do
+    revision 1
+    sha256 "6b3ae59be4da7b75d60b3fdaa75c6cbb9f602d96b343cb4233b59aa6ff4103ae" => :el_capitan
+    sha256 "bc93e42389818072869eac34d3d06985bce3d6d97048a2ae01e3f94296f535d2" => :yosemite
+    sha256 "10fd3a34aee2f907a4d4c71d694100c1e1df207fd6dd9a6c9b2ea50423330a17" => :mavericks
+  end
+
+  depends_on Camlp5TransitionalModeRequirement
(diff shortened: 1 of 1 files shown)


ilovezfs:
@mikemcquaid @apjanke 
The merge of https://github.com/Homebrew/homebrew/pull/48429 updating Coq to 8.5 actually broke CompCert in Homebrew master, which only has support for up to Coq 8.4, so hopefully this PR can be expedited since

- the formula is unchanged from its final 8.4 version in Homebrew master
- coq83 already exists here
- it is a dependency of Homebrew/homebrew#49490
- the only version of CompCert with 8.5 support is a third-party fork:
https://github.com/jhjourdan/CompCert/tree/coq8.5-compcert2.6

dunn:
We don't do cross-tap dependencies for core formulae though, so for CompCert it'll need to have this version vendored as a `resource`.

DomT4:
Aye, It should be vendored into the appropriate formula rather than submitted to Versions here, unless there's a reason to promise to support it for x period beyond being needed for one other formula.

ilovezfs:
Then why is coq83 in this repository?

ilovezfs:
Is an ENV.deparallelize scoped within a resource("foo").stage block or does it escape the block? If it escapes the block, what's the canonical way to restore the prior setting once the block returns? I ask because both Coq 8.4 in the PR and Coq 8.5 in master both build with ENV.j1, but compcert does not, and it would be nice not to have to deparallelize them both.

mikemcquaid:
> Then why is coq83 in this repository?

@ilovezfs For people who want to install just that version. We don't support using cross-tap versions dependencies in core as people may end up with versions they don't want/need and/or conflicts between formulae that depend on different versions.

ilovezfs:
@mikemcquaid That's fine. It just means that due to the Coq 8.5 upgrade, there will be significantly and gratuitously longer build times for CompCert since there won't be any intermediate bottles available for the build-time dependency, but so it goes. I imagine that's an issue all over the place at this point with the pervasive use of the practice of vendoring and no such thing as an intermediate bottle.

I've just resubmitted the CompCert PR with Coq 8.4 vendored in.

Feel free to close this unless anyone has separate interest in having a coq84 formula just as there is a coq83 formula here independent of the CompCert-speciific build dependency.

mikemcquaid:
> That's fine. It just means that due to the Coq 8.4->8.5 upgrade, there will be significantly and gratuitously longer build times for CompCert since there won't be any intermediate bottles available for the build-time dependency, but so it goes. I imagine that's an issue all over the place at this point with the pervasive use of the practice of vendoring and no such thing as an intermediate bottle.

Indeed, it's annoying and we're still working out a better solution for this.



> Feel free to close this unless anyone has separate interest in having a coq84 formula just as there is a coq83 formula here independent of the CompCert-speciific build dependency.

Thanks.

PR #1181, opened 2016-03-10
Add rebar 3.0.0 formula

rjcoelho:
Rebar3 is to much of a difference from rebar2  (available in homebrew/homebrew) and thus we need both simultaneous for some time


dunn:
I didn't realize it already existed in devel-only: https://github.com/Homebrew/homebrew-devel-only/blob/master/Formula/rebar3.rb; that's what's causing the audit failure.

Since there's a stable release now we can move it to here; if you could open an additional PR at devel-only that'd be great!

rjcoelho:
Do you want an additional PR at devel-only to remove rebar3 formula or to update it ?

dunn:
To remove it, thanks!

DomT4:
We could probably technically do a `tap_migration` :thought_balloon:

dunn:
But rebar3 was never in core?

DomT4:
I haven't tried this personally yet, but I _believe_ you can do tap to tap migration using a `.json` file. There's an example in what will become the new formula [repo](https://github.com/Homebrew/homebrew-core).

dunn:
Ohhh, cool.  I thought taps could only have `formula_renames` files.

rjcoelho:
Can someone explain how this tap migrations work ? And what do I need to do to. Thanks

rebar3.rb
```diff
@@ -0,0 +1,19 @@
+class Rebar3 < Formula
+  desc "Erlang build tool"
+  homepage "https://github.com/rebar/rebar3"
```
niku: It seems the repository has been moved from https://github.com/rebar/rebar3 to https://github.com/erlang/rebar3 .
So, homepage sholud be "https://github.com/erlang/rebar3".

mikemcquaid: :+1:

dunn:
@BrewTestBot test this please

dunn:
Sorry for the delay! Merged as fbb732e0bafc1b1acd9a59a73e61103582f0230f, thanks!

PR #1183, opened 2016-03-14
Add a formula for LLVM 3.8.0

chriso:
This is a copy of the existing llvm37 formula with some minor changes required for `3.8`.


llvm38.rb
```diff
@@ -0,0 +1,294 @@
...
+  test do
+    # test for sed errors since some llvm makefiles assume that sed
+    # understands '\n' which is true for gnu sed and not for bsd sed.
+    assert_no_match /PATH\)n/, (lib/"llvm-3.8/share/llvm/cmake/LLVMConfig.cmake").read
+    system "#{bin}/llvm-config-#{ver}", "--version"
+  end
+end
+
+__END__
+diff --git a/Makefile.rules b/Makefile.rules
```
chriso: This patch is still valid.

mikemcquaid:
Install is failing: http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1264/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_homebrew_versions_llvm38/

Did it work for you locally?

jenshnielsen:
The autoconf build system has been deprecated in [3.8](http://llvm.org/releases/3.8.0/docs/ReleaseNotes.html) so this seems as good a time as any to convert the formula to use CMake?

chriso:
@jenshnielsen the autoconf build system is deprecated in `3.8` and will be removed in `3.9`. It's quite a complicated homebrew build so I'm not sure if I'm the right person to make the switch.

@mikemcquaid it was mid-build when I submitted the PR - sorry about that. I'll fix it ASAP.

he-sk:
I could fix the install error with the following patch:

```patch
diff --git a/Library/Formula/llvm.rb b/Library/Formula/llvm.rb
index 9ff0156..e4b33f1 100644
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -205,15 +219,9 @@ class Llvm < Formula
 
     if build.with? "clang"
       (share/"clang/tools").install Dir["tools/clang/tools/scan-{build,view}"]
-      if build.head?
-        inreplace "#{share}/clang/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", "#{bin}/clang"
-        bin.install_symlink share/"clang/tools/scan-build/bin/scan-build", share/"clang/tools/scan-view/bin/scan-view"
-        man1.install_symlink share/"clang/tools/scan-build/man/scan-build.1"
-      else
-        inreplace "#{share}/clang/tools/scan-build/scan-build", "$RealBin/bin/clang", "#{bin}/clang"
-        bin.install_symlink share/"clang/tools/scan-build/scan-build", share/"clang/tools/scan-view/scan-view"
-        man1.install_symlink share/"clang/tools/scan-build/scan-build.1"
-      end
+      inreplace "#{share}/clang/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", "#{bin}/clang"
+      bin.install_symlink share/"clang/tools/scan-build/bin/scan-build", share/"clang/tools/scan-view/bin/scan-view"
+      man1.install_symlink share/"clang/tools/scan-build/man/scan-build.1"
     end
 
     # install llvm python bindings
```

chriso:
This should be good to go.

apjanke:
Looks good now. Merged.

Thank you for your contribution to Homebrew!

For future reference, we prefer to version bumps to be in a single commit, with comment formatted "formulaname X.Y.Z". I squashed the commits when pulling here, but you can squash them yourself when submitting by doing `git rebase -i master`, and the `git push --force <your-remote-name>`.

apjanke:
Autoclose didn't work because I manually squashed the commits on pull.

Closed in b39e77d2dc7a534aabb3a21e62553f77c9a90afe.

chriso:
@apjanke good to know. Thanks!

DomT4:
> The autoconf build system has been deprecated in 3.8 so this seems as good a time as any to convert the formula to use CMake?

We've already switched in the core. Just need some poor willing soul to volunteer to deal with all the Versions LLVM that should ideally be using CMake instead now :smile_cat:. If someone wanted to do it the core `llvm` formula provides a good base to work off of.

he-sk:
What needs to be done to update llvm to 3.8 in core?

DomT4:
Everything else that uses it needs to be compatible, which isn't the case yet. Our hands are tied there by various upstreams, sadly. In the nearish future we'll have a slightly more elegant solution but not there yet.

PR #1189, opened 2016-03-20
jpeg9 9b

s172262:


s172262:
I think that the reason of test failed may be that it has the new patch.

mikemcquaid:
That test failure is a Jenkins bug; don't worry about it.

jpeg9.rb
```diff
@@ -1,27 +1,27 @@
...
-    sha256 "be2a94a1715e16a4097ac00c7d9b5f3d4ebc64e96027bc4a40b321bb5da27eea" => :mountain_lion
   end
 
   option :universal
 
-  # https://trac.macports.org/ticket/37984
-  patch :DATA
+  stable do
+    # http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/graphics/jpeg/patches/patch-jmorecfg.h
+    patch :p0, :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

s172262: There are reports that this libjpeg header's bug causes the errors in some of the product depending on libjpeg (lcms, lcms2 and openmotif). Please see https://trac.macports.org/ticket/37984 .
But Mac Ports's patch which was included in previous version is rejected against 9b. (Please see the tests result of second commit in this branch.)
I think that the new patch is targeted at NetBSD originally but also it's useful on Mac OSX for the same goals of previous patch.

mikemcquaid: Sorry, I'm confused, does that mean it wasn't submitted upstream?

s172262: Is "upstream" the meaning as  Homebrew/_Homebrew_ ?
I tried _pull request_ of this script in  Homebrew/_Homebrew_. But all checks of dependence failed.
Because the formulas of depending on libjpeg has linked libjpeg**8**.dylib. This script makes only libjpeg**9**.dylib. 
So it isn't possible to replace  Homebrew/_Homebrew_'s jpeg.rb.
But this script is able to supersede homebrew-versions's jpeg9.rb of current version.

mikemcquaid: Sorry, upstream means the `jpeg` project at http://www.ijg.org

s172262: Sorry, I could not find out the feedback site of libjpeg.

s172262:
I close this PR. Because this makes the following error when building other code depended this.
> libtool:   error: '/usr/local/Cellar/jpeg9/9b/lib/libjpeg.la' is not a valid libtool archive

PR #1191, opened 2016-03-21
Add previous versions from qt5 series

scztt:
Added the largest dot version for each minor version from recent history. Copied directly from homebrew/Library/Formula/qt5.rb


mikemcquaid:
These should be named e.g `qt55` and `qt54` and only be the two prior versions before the current one. Also, is it necessary to have more than one version or do you just need 5.5?

DomT4:
Being somewhat blunt on this, I don't really want more than `qt55` here. The Qt5 formulae are an absolute nightmare to look after, deal with on CI and deal with on bugs. In an ideal world we wouldn't even have `qt55` here but I'll accept the necessity of it for now.

scztt:
I don't have a big stake in <5.5, I added those because it was easy to do at the same time. I can pull them out if there's consensus, and deal with it later if I end up needing. I updated the pr.

Qt removed a major set of packages and functionality in 5.6 (WebKit), so it's effectively more of a major-version change than a minor-version. There's really no good solution to this.... homebrew seems to assume non-major-version changes are generally backwards compatible, and in this case Qt 5.6 is not.

scztt:
If I need to do anything re Jenkins failures, please let me know - it's not clear to me whether it should be passing for pull requests.

qt55.rb
```diff
@@ -0,0 +1,194 @@
...
+  end
+end
+
+# Patches for Qt5 must be at the very least submitted to Qt's Gerrit codereview
+# rather than their bug-report Jira. The latter is rarely reviewed by Qt.
+class Qt5 < Formula
+  desc "Version 5 of the Qt framework"
+  homepage "https://www.qt.io/"
+  head "https://code.qt.io/qt/qt5.git", :branch => "5.5", :shallow => false
+  revision 2
```
DomT4: Don't need this. It's basically a new formula so:

* `revision`s get dropped.
* `deprecated_option`s get dropped.

qt55.rb
```diff
@@ -0,0 +1,194 @@
...
+  def message; <<-EOS.undent
+      To use --with-oci you have to set the ORACLE_HOME environment variable.
+      Check Oracle Instant Client documentation for more information.
+    EOS
+  end
+end
+
+# Patches for Qt5 must be at the very least submitted to Qt's Gerrit codereview
+# rather than their bug-report Jira. The latter is rarely reviewed by Qt.
+class Qt5 < Formula
```
DomT4: Jenkins is currently failing because you've not updated the `class` to match the filename:

```
Error: No available formula with the name "homebrew/versions/qt55" 
```

DomT4:
> Qt removed a major set of packages and functionality in 5.6 (WebKit), so it's effectively more of a major-version change than a minor-version. There's really no good solution to this.... homebrew seems to assume non-major-version changes are generally backwards compatible, and in this case Qt 5.6 is not.

This is something they'd been warning about for a fair while now, it shouldn't have surprised the amount of people it seems to have done. On Qt I'm less sure it's a Homebrew-moves-quickly thing than it is a Qt-ecosystem-moves-slowly thing, as much as I appreciate Homebrew is more of a "Debian Sid" than a "Debian Jessie".

We also don't assume everything is backwards compatible, and we're trying to improve the strength of tests in formula to check these things but previously when we've tried to keep Qt5 pegged in a relatively stable place we've been whomped with complaints that we weren't getting the latest Qt5 out to people fast enough. If you wanted to help rewrite tests on `qt5`-using formulae to help us catch incompatibilities better we'd more than welcome that.

scztt:
I've had no big qt5+homebrew problems like this for the qt5 series - this seems like a genuine outlier. I'd say it's a poor decision on the part of Qt, but I don't have enough context (though a quick github search shows that 98% of code is still using the now-removed WebKit packages). It's partly my fault for trying to use homebrew for automated provisioning in the way that you'd, say, use NPM to pull compatible packages. :) But there's no other good option, unfortunately.

DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/32342022db98e0414bf6432501e42f5093ee185b. Thank you for your contribution to Homebrew @scztt; we appreciate it!

PR #1201, opened 2016-04-01
node 4.4.2, 0.12.13, 0.10.44 + npm 2.15.1

chrmoritz:
Upgrades all node LTS / maintenance releases to the latest versions including a npm bump to 2.15.1 in all cases fixing a security issue: https://nodejs.org/en/blog/vulnerability/npm-tokens-leak-march-2016/


mikemcquaid:
Thanks again @chrmoritz!

PR #1202, opened 2016-04-04
postgresql93.rb: Update from version 9.3.11 to 9.3.12

alanorth:
Upstream released [9.3.12](http://www.postgresql.org/docs/9.5/static/release-9-3-12.html). Tested and working on Mac OS X 10.11.4.


mikemcquaid:

For future reference the preferred commit message format for simple version updates is `postgresql93 9.3.12`. Also, please don't sign-off your own commits. Cheers!

alanorth:
Ah, thanks for the tip about signing the commits. I instinctively use `git commit -S -s` when really, my intent was only to GPG sign. I will use `-S` only in the future. Thanks.

PR #1205, opened 2016-04-05
add docker 1.9.1

crhan:


mikemcquaid:
A few things that should be cleaned up at the same time as this:
- remove all but two docker versions (e.g. all but this one and `docker171`
- change the formula class name
- Consider making this called `Docker19` instead of `Docker191` so we can track patch versions

docker19.rb
```diff
@@ -0,0 +1,43 @@
+class Docker19 < Formula
+  desc "Pack, ship and run any application as a lightweight container"
+  homepage "https://www.docker.com/"
+  url "https://github.com/docker/docker.git",
+      :tag => "v1.9.1",
+      :revision => "a34a1d598c6096ed8b5ce5219e77d68e5cd85462"
+  revision 1
```
DomT4: Shouldn't have a `revision` on a "new formula", which this essentially is.

crhan: sorry, I'm not familiar with the formula. This code is just extract from here: https://github.com/Homebrew/legacy-homebrew/blob/773b115a696516b51d0be261dda4a7437e9f61f6/Library/Formula/docker.rb

DomT4: Sorry, I meant `revision 1` rather than `:revision => "a34a1d598c6096ed8b5ce5219e77d68e5cd85462"`. Apologies if that was unclear.

crhan:
why is that build task failed?

mikemcquaid:
@crhan http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1316/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_versions_docker19___strict___online/

apjanke:
Closing due to inactivity.

PR #1207, opened 2016-04-05
Adding ElasticSearch 1.5.2

henrytseng:
Amazon AWS only supports ElasticSearch 1.5.x currently so this will probably be convenient for several developers.

### DIFF
--- a/elasticsearch15.rb
+++ b/elasticsearch15.rb
@@ -1,0 +1,112 @@
+class Elasticsearch15 < Formula
+  desc "Distributed search & analytics engine"
+  homepage "https://www.elastic.co/products/elasticsearch"
+  url "https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.5.2.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Is this still supported upstream @jasontedor?

jasontedor:
> Is this still supported upstream

@DomT4 This version is no longer maintained by Elastic. Further, the AWS Elasticsearch service does not support all of the open source Elasticsearch APIs.

DomT4:
Thanks for the quick reply Jason. Passing on this then, I don't really want to support versions upstream no longer does on this.

@henrytseng If you wish to support this, since it doesn't need bottling you can get the same ease-of-access by setting up [your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). Thank you for your submission to Homebrew nonetheless!

henrytseng:
Okay!  Welp, if anyone is thinking about using it.  It is here: https://github.com/henrytseng/homebrew-elasticsearch

PR #1210, opened 2016-04-05
go 1.2, 1.3, 1.4: enable superenv shims

mistydemeo:
Since this isn't using a typical build tool like `make`, compiler argument refurnishment wasn't on. This led to build failures in at least go13 on recent versions of clang.


mikemcquaid:
> Since this isn't using a typical build tool like make, compiler argument refurnishment wasn't on.

Maybe we should add more tools e.g. the `go` build tools to this?

mistydemeo:
I'm not sure we can; it works for other tools because we wrap system commands like `make` (and set the environment variable there before calling the real tool), but here we're executing a shell script directly. It's not being executed via a wrapper.

mikemcquaid:
@mistydemeo 👍. I do wonder what it would be like if we always did compiler argument refurbishment _except_ for e.g. `configure` or cases we know it breaks.

PR #1216, opened 2016-04-10
New Formula: allegro 5.2.0

s172262:


DomT4:
What's the justification here? It seems to be a newer version of the one in `homebrew/core`?

s172262:
It may be as you say. Ver.5.2 release news say
> Allegro 5.2 is source but not binary compatible with 5.0.
> Allegro 5.2 is developed a little differently than 5.0 and 5.1 were. 

If you close this, I will move to `homebrew/core`.

DomT4:
Can you elaborate on both those points? Is this a development release or a stable release that's not backwards compatible with the 5.0.11 branch?

s172262:
Ver.5.2 release news say
> After many years of development in the unstable branch, the next stable version of Allegro is ready for release! This version brings most of the features from the unstable branch, but now with a guarantee of backwards source and ABI compatibility. Allegro 5.2 is source but not binary compatible with 5.0.
> Allegro 5.2 is developed a little differently than 5.0 and 5.1 were. This version combines both unstable and stable APIs in one branch. The unstable branch is no more! Unstable APIs may be changed between versions, or even removed entirely. 

They say this is stable and removed unstable code of ver.5.1 which has stopped to develop and unreleased in the end. 5.2 may be compatible with 5.0 but not 5.1. It can't be judged about incompatible with 5.1.
Current allegro.rb has `devel` part for 5.1. It may be self-responsibility to use `devel`. But there was the reason that 5.1 supported theora and libogg.

mikemcquaid:
Nothing in Homebrew depends on `allergro` so this should just replace the existing Homebrew `allegro` formula.

s172262:
I move this to `homebrew-core`. https://github.com/Homebrew/homebrew-core/pull/375

PR #1217, opened 2016-04-13
node4-lts 4.4.3

chrmoritz:
This PR
* bumps `node4-lts` to 4.4.3: https://nodejs.org/en/blog/release/v4.4.3/
* prevents head builds from acting like a stable version and
* makes the native addon test more robust for user of nvm by unsetting the `NVM_NODEJS_ORG_MIRROR`environment variable (deprecated in node-gyp 3, will be removed in v4)


DomT4:
Merged in https://github.com/Homebrew/homebrew-versions/commit/870b1daf58bb5d1dc9be16f6a4209ea196d09889. Thanks again @chrmoritz! I'll try and actually sit down and catch up with your core Node PR soon.

chrmoritz:
@DomT4: If you want to catch up there you probably want to start reading from this https://github.com/Homebrew/brew/pull/37#issuecomment-208381710. We've discarded the first design from the starting post because it was to resource heavy and are concentrating on fixing the major cache issue and changing the install format from local to global (with custom prefix to `libexec`) while sticking with npm based installs. Unfortunately this won't bring us offline installs as in first proposed design because it doesn't work well enough with homebrews current resource handling.

PR #1221, opened 2016-04-15
Add sip 4.17

m-kuhn:


DomT4:
We don't really do point releases here any more. Why would you want `4.17` specifically instead of `4.18`?

m-kuhn:
Our CI builds are broken since the update to 2.18 https://travis-ci.org/qgis/QGIS/builds/123309494
In the long run we will need to figure out how to fix it from our side but for now it would be good to be able to re-enable them.

mikemcquaid:
Sorry, we'd rather not include patch releases in this form. Please consider keeping this in your own tap. Sorry!

m-kuhn:
No worries, we're using it from the own repo now. I was proposing it here just in case there's interest.

mikemcquaid:
@m-kuhn cool, thanks!

PR #1228, opened 2016-04-20
ruby21 2.1.9

Jerzerak:


ruby21.rb
```diff
@@ -97,7 +97,7 @@ def post_install
     end
 
     # Create the version-specific bindir used by rubygems
-    mkdir_p "#{rubygems_bindir}"
+    mkdir_p '#{rubygems_bindir}'
```
DomT4: No need to change this line.

Jerzerak: @DomT4 `brew audit --strict --online ruby21` was spitting out that error and it also differs from the `mkdir_p` that is styled with a single quotation on line 96. Wouldn't we want them styled in parity?

DomT4: The one on line 96 is actually a backtick, not a single quote.

```
brew audit --strict --online ruby21
```
Is likely complaining because the #{}` is unnecessary. You should be able to escape that by changing it to:
```ruby
rubygems_bindir.mkpath
```

Jerzerak: Well that's what I get for staring at 10 point font when I'm this tired

ruby21.rb
```diff
@@ -97,7 +97,7 @@ def post_install
     end
 
     # Create the version-specific bindir used by rubygems
-    mkdir_p "#{rubygems_bindir}"
+    mkdir_p "rubygems_bindir.mkpath"
```
DomT4: Apologies if I was unclear here. I meant you could replace this whole line with:
```ruby
rubygems_bindir.mkpath
```

`mkpath` is just shorthand for `mkdir_p`.

Jerzerak: I really need to sleep now I think...that was way too obvious

ruby21.rb
```diff
@@ -1,8 +1,8 @@
 class Ruby21 < Formula
   desc "Powerful, clean, object-oriented scripting language"
   homepage "https://www.ruby-lang.org/"
-  url "https://cache.ruby-lang.org/pub/ruby/2.1/ruby-2.1.8.tar.bz2"
-  sha256 "250d0b589cba97caddc86a28849365ad0d475539448cf76bbae93190985b3387"
+  url "https://cache.ruby-lang.org/pub/ruby/2.1/ruby-2.1.9.tar.bz2"
+  sha256 "034cb9c50676d2c09b3b6cf5c8003585acea05008d9a29fa737c54d52c1eb70c"
```
DomT4: To add to your fun, our CI disagrees with your checksum here 😄:

```
==> Downloading https://cache.ruby-lang.org/pub/ruby/2.1/ruby-2.1.9.tar.bz2
Warning: Formula reports different SHA256: 034cb9c50676d2c09b3b6cf5c8003585acea05008d9a29fa737c54d52c1eb70c
Downloaded to: /Library/Caches/Homebrew/ruby21-2.1.9.tar.bz2
SHA256: 4f21376aa11e09b499c3254bbd839e68e053c0d18e28d61c428a32347269036e
```

DomT4:
@mistydemeo Any idea why `postinstall` is failing here now? The change there is stylistic rather than functionality, but perhaps I'm missing a gap in my already very gappy Ruby knowledge. I think you wrote this section, hence the ping, but apologies if I'm pinging the wrong person 🙈.

DomT4:
Merged as part of https://github.com/Homebrew/homebrew-versions/pull/1245. Commit credit there is still your's. Thank you for your contribution to Homebrew @Jerzerak; we appreciate it! :smiley_cat:

PR #1229, opened 2016-04-20
add libevent 2.1.5-beta, this version is latest version of libevent.

luafan:
Because there is no 2.1.x stable version, and it has been released over one year, so it's the only choice of stable version of 2.1.x.


mikemcquaid:
Sorry, we don't want to include beta versions in here even if they'd been out for a while.

PR #1230, opened 2016-04-20
Add docker 1.9.1

sztupy:


docker191.rb
```diff
@@ -0,0 +1,42 @@
+class Docker191 < Formula
```
mikemcquaid: Please make this `docker19` and remove `docker171` in the same PR, thanks!

sztupy: Can we keep `docker171`? The client and host versions of docker has to match meaning if we don't have a way of easily switching versions we won't be able to manage the different versions of servers

mikemcquaid: > Versions formulae must not exceed +/-2 versions from the current stable releas

https://github.com/Homebrew/homebrew-versions#acceptable-formulae

If you need to manage more on your servers I suggest creating your own tap.

apjanke: Ping. Still interested in moving 1.7.1 to 1.9 here?

It does sound like a Docker-specific tap is the way to go, since Homebrew-versions isn't going to give you enough version history coverage.

Note that per current Homebrew-versions policy, we should be removing `docker171` regardless of whether `docker19` gets added, since the version window is tied to upstream's current stable release, not the number of versions in homebrew-versions.

> Versions formulae must not exceed +/-2 major/minor (not patch) versions from the current stable release.

sztupy: Unfortunately with docker the client and the server versions should match, meaning the best thing would be to have a complete version coverage somewhere. Apparently this won't be homebrew-versions, so we'll have to move away anyway.

MikeMcQuaid: Ok, thanks.

PR #1231, opened 2016-04-20
gcc48 and gcc49: add gfortran by default

tschoonj:
This is mostly for selfish reasons: I want to use gfortran in Travis-CI but I cannot wait for gcc to build from source as it will make the build time out.

Also, this brings these packages in line with gcc5 and gcc6

### DIFF
--- a/gcc48.rb
+++ b/gcc48.rb
@@ -48,7 +48,7 @@ class Gcc48 < Formula
     end
   end
 
-  option "with-fortran", "Build the gfortran compiler"
+  option "without-fortran", "Build without the gfortran compiler"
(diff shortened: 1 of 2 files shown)


mikemcquaid:
Seems fine, yep 👍

DomT4:
Seems reasonable. Shipped in https://github.com/Homebrew/homebrew-versions/commit/567b80ae44572d3310d5edfee42668d429a57f3f & https://github.com/Homebrew/homebrew-versions/commit/2f191bfbfff1e874a1ab59d38e6f3a5cb4674c6c. Thanks Tom!

tschoonj:
Thanks!

PR #1234, opened 2016-04-21
Remove play12 formula

dcrall:
With the addition of the play14 formula, delete play12 per @mikemcquaid's direction. From the play14 pull request: 

> Would be good to remove e.g. play12 so we only have the current stable +/-2 versions here.


DomT4:
👍 Will merge this shortly. Can you make sure the:

```ruby
conflicts_with "play12", :because => "Both install a `play` executable"
```
Line is removed from the `play13` formula first though? Thanks!

DomT4:
And last but not least 🚢'd in https://github.com/Homebrew/homebrew-versions/commit/63bb8471eb89b8afe159ea9e317890625a26c2b3. Thank you for your efforts in getting this done @dcrall; we appreciate them ❤️.

PR #1235, opened 2016-04-22
Add Solr 5 recipe

magnusvk:
Which installs Solr 5.5 now that Solr 6 is released.

See also https://github.com/Homebrew/homebrew-core/pull/540

### DIFF
--- a/solr5.rb
+++ b/solr5.rb
@@ -1,0 +1,65 @@
+class Solr5 < Formula
+  desc "Enterprise search platform from the Apache Lucene project"
+  homepage "https://lucene.apache.org/solr/"
+  url "https://www.apache.org/dyn/closer.cgi?path=lucene/solr/5.5.0/solr-5.5.0.tgz"
+  mirror "https://archive.apache.org/dist/lucene/solr/5.5.0/solr-5.5.0.tgz"
+  sha256 "c5fa5cb996fe1432e09bb2f6053ffbeb095436db4a77e9c6488b531db726b04d"
+
+  bottle :unneeded
+
+  depends_on :java
+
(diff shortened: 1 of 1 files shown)


DomT4:
Can you kill off `solr14` and `solr36` at the same time (but with separate commits)? Thanks!

magnusvk:
Done.

magnusvk:
Jenkins test this please

dunn:
5.5.1 is out and apparently they don't continue mirroring old versions; we might want to do that ourselves on Bintray.

MikeMcQuaid:
I think if they don't mirror old versions we should probably close this unfortunately.

dunn:
@MikeMcQuaid they only mirror the latest release of the 5.x and the 6.x branch, so the core formula has the same problem.

magnusvk:
@dunn That's not true. [There is no 5.5.1.](http://archive.apache.org/dist/lucene/solr/) All the old versions remain mirrored without any problems. In fact, somebody else already [added a recipe for Solr 5.5.](https://github.com/Homebrew/homebrew-versions/blob/master/solr55.rb) Closing this.

dunn:
?? http://archive.apache.org/dist/lucene/solr/5.5.1/

magnusvk:
Funky. I must be blind. In any case, all the old files are still there also: http://archive.apache.org/dist/lucene/solr/5.5.0/

dunn:
Yeah, at http://archive.apache.org, but not at the mirrors: https://www.apache.org/dyn/closer.cgi?path=lucene/solr/5.5.0/solr-5.5.0.tgz (all those return 404).

PR #1247, opened 2016-05-05
Utvideo14.2.1 (New formula)

s172262:
Utvideo 15.0.0 and later couldn't link with ffmpeg.(See umezawatakeshi/utvideo#1 (in Japanese) or https://trac.ffmpeg.org/ticket/5511)
So `14.2.1` is last version to link with ffmpeg today.

$ otool -L /usr/local/Cellar/ffmpeg/3.0.2/lib/libavutil.55.17.103.dylib |sed -n -e "/(/s/(.*$//p" |sort -r
>	/usr/local/opt/zeromq/lib/libzmq.5.dylib 
	/usr/local/opt/x264/lib/libx264.148.dylib 
	/usr/local/opt/utvideo14/lib/libutvideo.14.2.1.dylib 
...
	/usr/lib/libc++.1.dylib 
...

And this do not provide QuickTime component. Please use newer version one.


mikemcquaid:
I think we'll pass on this as it doesn't have a proper homepage and the latest version isn't working. Sorry!

s172262:
If you can read Japanese, I`ll see URL is correct. 
And upstream say UtVideo is just a plug-in since 15.0.0. So we never build latest version to library.

mikemcquaid:
The URL is correct but it's not a dedicated homepage.

PR #1248, opened 2016-05-05
gnupg21 2.1.12

Midar:
Verified the SHA-256 using Werner Koch's public key, for which I
verified the fingerprint with Werner in person.


DomT4:
No need to remove the `bottle do` block FYI, CI automatically ignores and then regenerates it for jobs.

Merged in https://github.com/Homebrew/homebrew-versions/commit/d1232f4657e65a28cefb2d891e6034c70febcf75. Thank you for your contribution to Homebrew @Midar; we appreciate it! :smiley_cat:

PR #1250, opened 2016-05-05
Add Apache Thrift 0.9.2 recipe

hantonr:
Needed specific 0.9.2 version of Thrift. Only change from the brew core version of thrift is the url & sha256 hash + I added a test for homebrew audit and set to conflict with the main thrift version.

### DIFF
--- a/thrift092.rb
+++ b/thrift092.rb
@@ -1,0 +1,97 @@
+# This is the 0.9.2 version of the thrift formula and is identical
+# to the current 0.9.3 version besides the url and hash
+class Thrift092 < Formula
+  desc "Framework for scalable cross-language services development"
+  homepage "https://thrift.apache.org"
+
+  stable do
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Sorry, we're not supporting versions that vary only by patch version now. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

hantonr:
Thanks @mikemcquaid - that's a bummer that homebrew doesn't support patch versions directly. If everyone has to go make their own taps, are there any projects out there to go find all the taps/formula on github (i.e. maybe someone on some tap somewhere already made the formula I need)? Might be an interesting thing to do.

mikemcquaid:
@hantonr There is not really, sorry. If someone else maintained it you may find you have the same problems, though 😉

PR #1251, opened 2016-05-06
node4-lts 4.4.4, node5 5.11.1 (security updates)

chrmoritz:
Upgrading node4-lts and node5 to the latest releases fixing openssl vulnerabilities: https://nodejs.org/en/blog/vulnerability/openssl-may-2016/.
The node 0.10 and 0.12 are not ready yet, but will follow most likely later today.


node5.rb
```diff
@@ -1,17 +1,15 @@
...
+
   bottle do
     sha256 "591be9bf371ffefac664146b84e3517c189a7db7eae0dfd6cd45d5126c915cc1" => :el_capitan
     sha256 "9678fb9530633e9f3bf76daba68919a820b9baa663d3e11f715d898a444d4db8" => :yosemite
     sha256 "ca964f5738c01a64165bf059da4488524ff0ccc6d9ea59ef87c6b730d9603933" => :mavericks
   end
 
-  # uncomment once a v5.x staging branch is cut
-  # head "https://github.com/nodejs/node.git", :branch => "v5.x-staging"
-
```
chrmoritz: Looks like we don't get a staging branch for v5 (and using the `v5.x` branch here would be most the time just an alternative download strategy for stable, except immediately before/after a release).

chrmoritz:
BTW: With the next node v6 version (6.2?) node will ship with a pre-canned small ICU build in their source tarballs. This means that the icu resource will be no longer needed for the default small-icu build. You can already look at an (for now head-only) formula using this new icu build system over at: https://github.com/chrmoritz/homebrew-core/commit/b27dc90f93ccd20334e3a879ba7d41b5ea41d4d9 (tested with both small and full icu).

mikemcquaid:
Thanks again @chrmoritz!

DomT4:
> BTW: With the next node v6 version (6.2?) node will ship with a pre-canned small ICU build in their source tarballs. This means that the icu resource will be no longer needed for the default small-icu build.

This is nice to hear! It's a fairly hefty resource so minimising how frequently it is needed is certainly a plus.

PR #1252, opened 2016-05-06
README: note patch versions.

mikemcquaid:
@DomT4 Thoughts? Think this may avoid confusion and feels like a good policy if we don't have it already.


DomT4:
👍 :shipit:

PR #1253, opened 2016-05-06
node012 0.12.14

Jerzerak:


DomT4:
Does the internal `npm` need updating at all? The upstream Changelog should advise you if `npm` was updated by them.

chrmoritz:
> Does the internal npm need updating at all? The upstream Changelog should advise you if npm was updated by them.

Nope. Last time node 0.10/0.12 have updated their npm, they shipped with an npm version 2.15.1 which erroneous reported it's version as 2.15.0. We weren't affected by this, because we use our own `resource` for installing npm.
The commit in this release only fixes the wrong version number, which was never incorrect in our builds.

DomT4:
Thanks for the added detail @chrmoritz, appreciated as always.

Merged in https://github.com/Homebrew/homebrew-versions/commit/eec8db57ba8e9f1c77bd4acd29a1d889fc453494. Thank you once more @Jerzerak! 😺

PR #1256, opened 2016-05-09
gcc6: 6.1.0

manphiz:
Upgrade to latest stable version of GCC 6.

Note that the version number seems to be lower than the previous snapshot (6-20160320), so a force reinstall is needed. Wonder if there is a way to force upgrade to lower versions like epoch in other systems?

### DIFF
--- a/gcc6.rb
+++ b/gcc6.rb
@@ -21,8 +21,9 @@ class Gcc6 < Formula
 
   desc "The GNU Compiler Collection"
   homepage "https://gcc.gnu.org"
-  url "ftp://gcc.gnu.org/pub/gcc/snapshots/6-20160320/gcc-6-20160320.tar.bz2"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Wonder if there is a way to force upgrade to lower versions like epoch in other systems?

There is not but we should make one. PR (or even a well-written up issue) would be extremely welcome.

DomT4:
Yeah, we should have thought about this when we added it really. Could have hacked it by doing something like `version 0_20160320` possibly.

Way back in the day I filed [this](https://github.com/Homebrew/legacy-homebrew/issues/41563) on epochs.

mikemcquaid:
Thanks again @manphiz!

manphiz:
Thanks Mike!

Will figure out what to do with epoch later.

PR #1257, opened 2016-05-09
subversion 1.8.16

Jerzerak:


mikemcquaid:
Thanks again @Jerzerak!

PR #1259, opened 2016-05-11
Create R 3.2.5

JDeanThomas:
R 3.2.5 is a very stable build and it will be some time before many highly used packages are verified R 3.3.0 compliant or many users (especially in academia) are prepared to upgrade.

### DIFF
--- a/R325
+++ b/R325
@@ -1,0 +1,225 @@
+class RDownloadStrategy < SubversionDownloadStrategy
+  def stage
+    cp_r File.join(cached_location, "."), Dir.pwd
+  end
+end
+
+class R < Formula
+  desc "Software environment for statistical computing"
+  homepage "http://www.r-project.org/"
+  url "http://cran.rstudio.com/src/base/R-3/R-3.2.5.tar.gz"
+  mirror "http://cran.r-project.org/src/base/R-3/R-3.2.5.tar.gz"
+  sha256 "60745672dce5ddc201806fa59f6d4e0ba6554d8ed78d0f9f0d79a629978f80b5"
+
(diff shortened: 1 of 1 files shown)


DomT4:
Honestly, I'd rather @Homebrew/science keep Science versions. If anything breaks here they'll only get pinged any way because no Homebrew maintainer has a strong mutual interest in both Science and Versions, and this repo is _slowly_ going to be used for less and less as we move to an alternative system.

JDeanThomas:
@DomT4 Either makes sense to me, though I don't really see versions in /science.

I am sort of putting together a coherent and comprehensive guide for compiling R and the range of options/additions etc using Homebrew so that it's accessible to, say, graduate students to compile packages using GCC as well so that they can harness the full power of open tools like OpenMx with reduced learning curve. I'm hoping to add some additions like compiling intel's MKL and the full range of option from the R manual, so I'll reach out to the maintainer of the R formula. But my experience with R (and how researchers use it to make reproducibility as easy as possible) is to have the last build of the last major version pinned somewhere until it's moved two major versions forward, ie keeping a 3.2.5 until there is a 3.4. I don't think versions need be any deeper than that.

dpo:
I'm not so much in favor of versions in /science. The plethora of formulas is difficult to maintain already and PRs about R typically take days to weeks to go through.

@JDeanThomas You may want to consider setting up your own tap until 3.4 comes out.

JDeanThomas:
@dpo Sounds good. I'll setup a tap, for now centered around R and GCC and if there is demand/requests, I'll add older versions and maintain it.

PR #1269, opened 2016-05-24
node4-lts 4.4.5

chrmoritz:
Upgrades `node4-lts` to version 4.4.5: https://nodejs.org/en/blog/release/v4.4.5/


chrmoritz:
Looks something went wrong with the npm upgrade (looking into it now).

chrmoritz:
Working npm versions: 2.15.4 and below **and** 2.15.6 (atm unstable lts version)
Not working: npm 2.15.5 (only this version): https://github.com/npm/npm/releases/tag/v2.15.5

I guess there is something wrong with unofficial registry tarball of npm@2.15.5 we are using given that only this specific version (2.15.5) is affected and the upcoming 2.15.6 is working fine (and it does not contain any fix which could be related).

I think the best way to go is using npm 2.15.4 for now.

MikeMcQuaid:
> I think the best way to go is using npm 2.15.4 for now.

@chrmoritz Sounds good 👍

MikeMcQuaid:
Thanks again @chrmoritz!

PR #1271, opened 2016-05-28
Add ansible 2.0.2

powersjcb:
Pulled from Homebrew/homebrew-core: af9c040d4ba13d6c023d29c4d7326ab722768a49

Ansible 2.1 is new default.

PR #1274, opened 2016-05-30
Remove unacceptable versions

MikeMcQuaid:

-----

Remove "unacceptable" versions of formulae based on a loose interpretation of https://github.com/Homebrew/homebrew-versions#acceptable-formulae meaning we shouldn't have more than 2 versions "higher" or "lower" than the version in homebrew/core.

Other things I'd like to try and do to clean up this tap a bit:
- Add an `audit` rule to detect this in future
- Decide on a consistent versioning scheme for both here and homebrew-core versions e.g. based on major/minor versions (and patch versions only in exceptional circumstances)
- Possibly do a second pass and remove those formulae that don't meet the -/+ 2 versions more strictly based on major/minor versioning.

CC @DomT4 and @UniqMartin who have talked about this before.


DomT4:
I'd get rid of `subversion17` as well personally, given how much of a giant hack that formula is these days on the OpenSSL situation.

[This](https://gist.github.com/DomT4/37aa91a761c5b8aae81329dfb32ba3b9) is my current working file on killing stuff we no longer want to or can support FWIW. Doesn't seem to be anything I haven't agreed with here.

MikeMcQuaid:
@DomT4 Do you think we want to move to `versions-unsupported` or just remove?

DomT4:
I'm torn, to be honest. To make mass-deletion a bit softer on existing users we tend to move to the `boneyard` rather than kill and I think that system is sane. Maybe we could just boneyard them rather than creating a whole new tap for it or something.

MikeMcQuaid:
@DomT4 How do `tap_migrations` work between taps? I think rather than boneyarding the Git history is sufficient if people still need to be running these old versions.

DomT4:
@MikeMcQuaid I haven't independently verified this yet but AFAIK they're _supposed_ to work just like `homebrew/core` these days. Theoretically. My worry isn't so much supporting them but not causing loading errors if people have custom formulae in taps that `depends_on "go13"` or whatever.

UniqMartin:
I'm basically :+1: on the general idea and I don't feel particularly attached to any of the formulae being eliminated here (though that's probably not valuable information). I think we should do the usual tap migration dance and either have `boneyard` as the destination or `versions-unsupported` (which to me reads just like a specialized `boneyard` for `versions`). Or have we just deleted instead of boneyarded formulae in the past? (I guess I'm lacking some historical context here.)

As far as I understand the migration should also work for non-core taps, but I haven't tested this. I'm also not clear how this works in the hypothetical case when more than two taps are involved (if this is possible at all). I guess previously this was a bit easier when `core` wasn't just a tap (with privileges).

> - Add an `audit` rule to detect this in future

I guess that's a good idea, but I'm not sure yet how to make such a rule reliable. Maybe this ties immediately into the following. (And we'd still need some way for escaping this rule for things like `gcc`.)

> - Decide on a consistent versioning scheme for both here and homebrew-core versions e.g. based on major/minor versions (and patch versions only in exceptional circumstances)

:+1: though the exact scheme will obviously be dictated by upstream's policy on versioning their releases (including things like version scheme changes, e.g. `gcc` 4.9.x followed by 5.x, 6.x, …).

> - Possibly do a second pass and remove those formulae that don't meet the -/+ 2 versions more strictly based on major/minor versioning.

I think one criterion that could inform this second pass is whether something has an El Capitan bottle or whether adding one is just a matter of recompiling. If it breaks (horribly), it most likely should be removed.

ilovezfs:
> I think we should do the usual tap migration dance and either have boneyard as the destination or versions-unsupported (which to me reads just like a specialized boneyard for versions). Or have we just deleted instead of boneyarded formulae in the past? (I guess I'm lacking some historical context here.)

I think it would be better to handle deleted formulae gracefully rather than clutter boneyard.

DomT4:
> Or have we just deleted instead of boneyarded formulae in the past? 

Historically tap stuff just gets "taken out back and shot" as the expression goes, but for as long as I've been involved in Versions we've never wholesale removed on a scale even close to this before.

MikeMcQuaid:
Yeh, I should at least try the `tap_migrations` here.

MikeMcQuaid:
So, thinking about it more, we've been removing versions from stuff like https://github.com/Homebrew/homebrew-versions/pull/1271 somewhat regularly with basically no complaints. I see this PR as an extension of that. I think the boneyard is a little different as we're removing stuff then that's broken rather than old and unsupported (which additionally tends to imply "insecure"). If folks disagree strongly I'm happy to comply but I think it may be worth doing this as-is and we can always consider adding the `tap_migrations` if there's sufficient complaining or I'm wrong in my assumptions above.

ilovezfs:
I agree, as I said above. If outright removals cause problems (which it doesn't sound like they actually do), those are bug(s) that should be fixed, and aren't an excuse to abuse boneyard.

DomT4:
>  If outright removals cause problems (which it doesn't sound like they actually do), 

Add `depends_on "homebrew/versions/ansible14` to `wget` in Homebrew, cd into the Versions tap and `rm ansible14.rb` and then the fun begins:

```
brew install wget
Error: No available formula with the name "homebrew/versions/ansible14" (dependency of wget)
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

```
brew deps wget
Error: No available formula with the name "homebrew/versions/ansible14"
```

```
brew install -v wgetpaste
Error: No available formula with the name "homebrew/versions/ansible14" (dependency of wgetpaste)
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

It's not a major situation by any means, but we are likely to cause _some_ breakage that isn't entirely clear why or what happened to end users by en mass deletion.

MikeMcQuaid:
@DomT4 A fair point; I'll check which, if any, of these have dependents in the Homebrew org taps and have a general think about what people will `depends_on` (e.g. not `ansible`)

DomT4:
Maybe the answer for taps is `deleted_formulae.json` or something that causes a more helpful error message than essentially _"Erm, I seem to have lost this formula, oops"_. Not sure.

ilovezfs:
I don't think we need a running obituary. If it's not found, it's not found. The breakage is legitimate, and they should update accordingly.

DomT4:
Sure, but "go ahead and potentially shaft people" is pretty bad UX to be on the end of, especially if you encounter that breakage through a script or something like Travis where it's worked for years and suddenly stops doing so.

I'm in favour of all these dying or being moved elsewhere, couldn't be _more in favour_ of removing oh so much of Versions, even beyond this initial set; I just think our current system for deletion is inelegant and that makes me worry a bit in terms of a move designed to lessen our maintenance burden ends up temporarily pushing it upwards.

ilovezfs:
Yeah, the problem is once the obituary database becomes a thing, good luck getting rid of it.

MikeMcQuaid:
My concern is similar; if we make the migration too simple are we enabling (or actively encouraging) people to run insecure software?

DomT4:
I don't think there's a winning position here, in terms of security. Even if we remove them from Versions they remain on the user's machines and users aren't notified of our changes in terms of security, we don't throw warnings like:

```
brew doctor

* `ansible14` was removed due to security concerns. We recommend you uninstall it.
```
Instead it sits there silently indefinitely until the user investigates a fair bit. Our security handling for everything else is `brew update` "and upgrade the impacted formula", but that does nothing on stuff we actively remove because it could be or is full of holes you could fit an asteroid through.

I'm fine if you want to merge this as-is. My complaint is more general in terms of how badly we handle formula that vanish on our side from taps.

ilovezfs:
Do you know of an example of a package manager that handles it well?

DomT4:
Not off the top of my head.

Most of the Linux PMs just backport security fixes for years and years and years after software stops being supported upstream, but we don't have the humanpower to do anything like that. But I don't find "Homebrew is as awful as everyone else" a particularly motivating point to be honest.

I'm not advocating we go in and hard remove software from user's machines, but we could be a bit more transparent about removing things because they're potentially unsafe.

ilovezfs:
brew audit could report out a generic "unknown to any tap. possibly removed for security reasons" for anything that has an install receipt but is now gone.

DomT4:
`brew doctor` rather than `brew audit` presumably, given `brew audit` isn't something we ever encourage non-contributing users to engage with?

Do local formula installed, i.e. not from a tap, generate a receipt? I can't remember, has been a while. If they do that more or less makes the idea problematic pretty quickly, sadly.

ilovezfs:
>brew doctor rather than brew audit presumably,

Yes, see correction above.

They get a receipt, but
```
"tap":null,"path":"/Users/joe/wget2.rb"
```

DomT4:
I guess you could doctor for anything that isn't `tap:null` and doesn't have a parent formula I guess, but anything like that is going to be an extremely expensive operation to perform every single `doctor` run, particularly if you have a lot of formulae installed.

To be honest I think we just have to merge this and live with it. This is the devil we got in bed with & there's very few reasonable ways out that don't involve either not insignificant added logic to the core to handle tap obituaries or costly operations to go sniffing the filesystem for installed formulae with missing parents.

IMO, merge this as-is and let it be a lesson to us not to accept so much versioned crap we can't possibly keep on top of long-term.

UniqMartin:
I think merging it as-is is fine. By now I'm convinced that neither `tap_migrations.json` nor an obituary in the form of `deleted_formulae.json` is a really good solution here.

I do agree with @DomT4 that the UX for deleted (or even boneyarded formulae, to think a bit futher) could be improved though. We could try to give some indication that a formula previously available through an official tap is now no longer available or has been boneyarded, suggesting the user might want to remove it. `brew doctor` feels like the natural place for that, though something tucked underneath the `update-report` also seems like a good idea.

> I guess you could doctor for anything that isn't `tap:null` and doesn't have a parent formula, but anything like that is going to be an extremely expensive operation to perform every single `doctor` run, particularly if you have a lot of formulae installed.

I don't agree. Iterating over (at most) a few hundred `INSTALL_RECEIPT.json` files that are both very small and in a fixed location should be super quick and we do a lot more expensive operations in `brew doctor` (like traversing entire trees in search for broken symlinks). I'm confident we can easily do this both in `brew doctor` and (I'm getting increasingly convinced about this) at the end of `brew update` (in `update-report`).

UniqMartin:
> IMO, merge this as-is and let it be a lesson to us not to accept so much versioned crap we can't possibly keep on top of long-term.

😅 👍 Let's merge now and follow-up with the UX improvements in `doctor`/`update` later (if everyone agrees that's a good thing) as we've lived with this situation for a while already and can do so a bit longer.

MikeMcQuaid:
🚢d. Time to wait for 💥

zzatking:
If anyone needs a version that was removed due to this pull request, you can find it by following these steps:

1. Click the commit above related to your software (e.g. ``cassandra*``)
2. Find the formula you desire (e.g. ``cassandra12.rb``)
3. Click "View" for that formula
4. Click "Raw" for the source
5. Copy the URL
6. Run `brew install <URL>`

DomT4:
If people actually want to maintain them, and half the problem is that to date people have not, setting up your [own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) is the ideal mechanism there.

dideler:
> If people actually want to maintain them

Genuine question as someone affected by this change, could you elaborate on the maintenance required? Is the main concern keeping the patch level up to date for versioned formulae?

DomT4:
I'll touch on a few things in this reply to preempt other questions people may have on motivation, so if you want the tl;dr it's the bullet list below.

Outside of Homebrew official, you aren't bound by our rules. Support it as much as you want to support it, and if your tap gathers steam and community popularity we're happy to consider nudging people towards if it they want for example an old old version of `cmake`, etc. 

Homebrew's needs/wants/concerns/questions/etc primarily include but aren't necessarily limited to:

* Ensuring the formula actually builds, period 😄.
* Ensuring the formula actually builds on an OS X version we can test, and consequently providing CI time & space for these formulae to be tested as necessary.
* Keeping bottles fairly up-to-date and keeping those bottles hosted indefinitely.
* Ensuring the version is as up-to-date as it can be.
* When security flaws are discovered, do we endlessly backport the fixes or kill the formula? Due to having a fairly small team the answer tends to be option B there. Are we even familiar enough with that particular piece of software to safely and responsibly backport security fixes for it?
* When upstreams declare they no longer support this particular version and will ignore bug reports for it, how do we handle it when we start getting bug reports given the small team & lack of intimacy with every single piece of software Homebrew packages?
* If a formula is getting 70 downloads a year, is it worth the support burden it takes given older versions by their nature have a higher support burden than newer, upstream-supported releases which are more widely used within the Homebrew ecosystem?
* Is it responsible for a package manager to package known insecure versions in general availability?
* Keeping the style usage somewhat in step with the core so contributors find cross-contributions easier and more consistent.
* Backporting tests from core to the Versioned formulae so we know the libraries/executables are functioning as expected, as best we can.

Essentially the problem is that everyone chucked everything they wanted for project xyz or personal use at Versions and we said yes to most of them without much consideration of _"So, what do we do in x years when this isn't supported by upstream and people still expect us to make it work"_. That was a mistake, on our part, evidently.

If you look at the contribution graph from the end of [2014 till today](https://github.com/Homebrew/homebrew-versions/graphs/contributors?from=2014-12-29&to=2016-06-02&type=c) the highest number of commits from anyone that isn't or hasn't been a maintainer of the project is 25, total, over that year and a half or so, and even that is _high_ for Versions in terms of non-maintainer commits. There's also a lot here that Bintray's download numbers show isn't even downloaded once a day.

Homebrew is trying to get to a place where our versions support is a lot more consistent, and well, safe for end users, but to get there the old cruft has to be peeled away.

zzatking:
In my opinion, it's better that Homebrew doesn't support super old versions of software. In my case, I had to install an old piece of software that is now considered an "archive version" on the author's downloads page. That goes to show you the lack of support there is even from the author of the software itself.

If the infrastructure I work with depends on an "archived version" of software, it's not the difficulties of installing the older piece of software that I should be dealing with (which includes bothering the Homebrew folks like you, @DomT4), it is the difficulties of nudging those in charge of the infrastructure to get them to upgrade the dependent version. 

here are few exceptions to avoiding software upgrades (e.g. Python 2 to Python 3), and generally upgrading software leads to ease of use, more security, and a potentially better workflow.

MikeMcQuaid:
> If the infrastructure I work with depends on an "archived version" of software, it's not the difficulties of installing the older piece of software that I should be dealing with (which includes bothering the Homebrew folks like you, @DomT4), it is the difficulties of nudging those in charge of the infrastructure to get them to upgrade the dependent version that I should be concerned with.

Great points @zzatking 👍

randy3k:
RStudio is still using boost150, I am afraid that removing boost150 will break  RStudio Server https://github.com/Homebrew/homebrew-science/pull/3663

Where should a old formula go if there is a real need of that formula?

MikeMcQuaid:
@randy3k It can either vendor it as a `resource` or should be removed if it depends on old, unmaintained software.

i-am-the-slime:
Hey guys, I need maven version 3.0. Now you removed it. What do I do? Isn't this the whole point of homebrew versions to actually be able to get this old *unacceptable* sh!t?

dideler:
@i-am-the-slime there's a couple of options to install the formula of a removed version.

As a one-off, you can `brew install <github-url-to-historic-file>` as described in https://github.com/Homebrew/homebrew-versions/pull/1274#issuecomment-223403003

For a more permanent solution, you can use a [tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md).

MikeMcQuaid:
> For a more permanent solution, you can use a tap.

This is what we recommend. To elaborate further: you can copy the file we removed here into your own tap and maintain it yourself. Thanks!

DomT4:
>  Isn't this the whole point of homebrew versions to actually be able to get this old unacceptable sh!t?

To explicitly answer this question, no. Versions is supposed to be a place for testable, recent older versions that Homebrew can reasonably support with CI, bottles & testing. It was never intended to be a dumping ground for absolutely everything and anything, even if that's what it became.

joshka:
Back reference from http://stackoverflow.com/questions/38061618/brew-install-cassandra12-fails-on-mac

PR #1316, opened 2016-07-18
gnupg21 2.1.15

ilovezfs:
- remove upstreamed patch
- remove head spec
- only run `make check` optionally

### DIFF
--- a/gnupg21.rb
+++ b/gnupg21.rb
@@ -1,9 +1,9 @@
 class Gnupg21 < Formula
   desc "GNU Privacy Guard: a free PGP replacement"
   homepage "https://www.gnupg.org/"
-  url "https://gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.13.tar.bz2"
-  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gnupg/gnupg-2.1.13.tar.bz2"
-  sha256 "4f9d83a6221daa60130fa79f0b1d37d6c20fffdd0320b640c7a597c5b6219675"
+  url "https://gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.15.tar.bz2"
+  mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/gnupg/gnupg-2.1.15.tar.bz2"
+  sha256 "c28c1a208f1b8ad63bdb6b88d252f6734ff4d33de6b54e38494b11d49e00ffdd"
 
   bottle do
     sha256 "4badc6179f850d075d1068e1b5dd2cb9e8c57a58f1a21ffc561aa755bfddf945" => :el_capitan
@@ -11,15 +11,8 @@ class Gnupg21 < Formula
     sha256 "e759f7df6242fd5f25bb18e7c9daa880cfb6f56bcddc067ff37dfbf1d5f65e0d" => :mavericks
   end
 
-  head do
-    url "git://git.gnupg.org/gnupg.git"
-
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-  end
-
   option "with-gpgsplit", "Additionally install the gpgsplit utility"
+  option "with-test", "Verify the build with `make check`"
 
   depends_on "pkg-config" => :build
   depends_on "sqlite" => :build if MacOS.version == :mavericks
@@ -37,26 +30,17 @@ class Gnupg21 < Formula
   depends_on "homebrew/fuse/encfs" => :optional
 
   conflicts_with "gnupg2",
-        :because => "GPG2.1.x is incompatible with the 2.0.x branch."
+        because: "GPG2.1.x is incompatible with the 2.0.x branch."
   conflicts_with "gpg-agent",
-        :because => "GPG2.1.x ships an internal gpg-agent which it must use."
+        because: "GPG2.1.x ships an internal gpg-agent which it must use."
   conflicts_with "dirmngr",
-        :because => "GPG2.1.x ships an internal dirmngr which it it must use."
+        because: "GPG2.1.x ships an internal dirmngr which it it must use."
   conflicts_with "fwknop",
-        :because => "fwknop expects to use a `gpgme` with Homebrew/Homebrew's gnupg2."
+        because: "fwknop expects to use a `gpgme` with Homebrew/Homebrew's gnupg2."
   conflicts_with "gpgme",
-        :because => "gpgme currently requires 1.x.x or 2.0.x."
-
-  # Patch for `make check` failures in 2.1.13.
-  # Fixed upstream by Justus Winter & can be removed in next release.
-  patch do
-    url "https://raw.githubusercontent.com/Homebrew/formula-patches/7b2211b/gnupg21/spawned_child_8f79c31b.diff"
-    sha256 "fde57b1e484aa738af8f582910330cadef144430d9221d3242b5f635c7f41dbe"
-  end
+        because: "gpgme currently requires 1.x.x or 2.0.x."
 
   def install
-    (var/"run").mkpath
-
     ENV.append "LDFLAGS", "-lresolv"
     ENV["gl_cv_absolute_stdint_h"] = "#{MacOS.sdk_path}/usr/include/stdint.h"
 
@@ -72,12 +56,6 @@ class Gnupg21 < Formula
 
     args << "--with-readline=#{Formula["readline"].opt_prefix}" if build.with? "readline"
 
-    if build.head?
-      args << "--enable-maintainer-mode"
-      system "./autogen.sh", "--force"
-      system "automake", "--add-missing"
-    end
-
     # Adjust package name to fit our scheme of packaging both gnupg 1.x and
     # and 2.1.x and gpg-agent separately.
     inreplace "configure" do |s|
@@ -88,7 +66,11 @@ class Gnupg21 < Formula
     system "./configure", *args
 
     system "make"
-    system "make", "check"
+
+    # intermittent "FAIL: gpgtar.scm" and "FAIL: ssh.scm"
+    # https://bugs.gnupg.org/gnupg/issue2425
+    system "make", "check" if build.with? "test"
+
     system "make", "install"
 
     bin.install "tools/gpgsplit" => "gpgsplit2" if build.with? "gpgsplit"
@@ -98,6 +80,10 @@ class Gnupg21 < Formula
     mv share/"doc/gnupg2/examples/gpgconf.conf", share/"doc/gnupg2/examples/gpgconf21.conf"
     mv share/"info/gnupg.info", share/"info/gnupg21.info"
     mv man7/"gnupg.7", man7/"gnupg21.7"
+  end
+
+  def post_install
+    (var/"run").mkpath
   end
 
   def caveats; <<-EOS.undent
@@ -122,6 +108,6 @@ class Gnupg21 < Formula
   end
 
   test do
-    system "#{bin}/gpgconf"
+    system bin/"gpgconf"
   end
 end


ilovezfs:
needs https://github.com/Homebrew/homebrew-core/pull/3101
needs https://github.com/Homebrew/homebrew-core/pull/3102

MikeMcQuaid:
👍 when ready

ilovezfs:
Two fixes work here: either a) drop `depends_on gettext` or b) make this change
```
-    ENV.append "LDFLAGS", "-lresolv"
+    ENV.append "LDFLAGS", "-lresolv -lintl"
```

@DomT4 @MikeMcQuaid got a preference?

Jenkins error is
```
Undefined symbols for architecture x86_64:
  "_libintl_bind_textdomain_codeset", referenced from:
      _i18n_switchto_utf8 in libcommon.a(libcommon_a-i18n.o)
      _i18n_switchback in libcommon.a(libcommon_a-i18n.o)
  "_libintl_bindtextdomain", referenced from:
      _i18n_init in libcommon.a(libcommon_a-i18n.o)
      _i18n_localegettext in libcommon.a(libcommon_a-i18n.o)
  "_libintl_gettext", referenced from:
      _my_strusage in gpgscm-main.o
      _create_pipe_and_estream in libcommon.a(libcommon_a-exechelp-posix.o)
      _gnupg_spawn_process in libcommon.a(libcommon_a-exechelp-posix.o)
      _gnupg_spawn_process_fd in libcommon.a(libcommon_a-exechelp-posix.o)
      _gnupg_wait_processes in libcommon.a(libcommon_a-exechelp-posix.o)
      _gnupg_spawn_process_detached in libcommon.a(libcommon_a-exechelp-posix.o)
      _disable_core_dumps in libcommon.a(libcommon_a-sysutils.o)
      ...
  "_libintl_setlocale", referenced from:
      _i18n_init in libcommon.a(libcommon_a-i18n.o)
      _i18n_localegettext in libcommon.a(libcommon_a-i18n.o)
  "_libintl_textdomain", referenced from:
      _i18n_init in libcommon.a(libcommon_a-i18n.o)
ld: symbol(s) not found for architecture x86_64
```

DomT4:
Option B seems fine, but would be nice to know what they changed that necessitated that.

ilovezfs:
@DomT4 safe bet: someone "broke" the configure.ac or Makefile.am!

ilovezfs:
@DomT4 Ug and the HEAD build is broken :(

```
/usr/local/opt/gettext/bin/msgfmt: found 866 fatal errors
2174 translated messages.
make[3]: *** [en@quot.gmo] Error 1
make[2]: *** [stamp-po] Error 2
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

DomT4:
And people keep asking why I won't accept GnuPG 2.1.x to replace the `gnupg2` formula in the core yet `¯\_(ツ)_/¯`.

Maybe report both those issues upstream & relegate NLS to being optional for now then.

ilovezfs:
I'm leaning toward adding `--without-disappearing-ink` and `--without-secret-decoder-ring` as options.

DomT4:
I'm not even sure if you're pulling my leg with those being real configure flags 😅

DomT4:
Looks like upstream realised the error with `gettext`'s linking at least: https://github.com/Homebrew/formula-patches/pull/43

ilovezfs:
Maybe we should wait for them to cut a release since that's a fairly large blunder.

DomT4:
Releases are monthly at this point. I suspect if we try to wait we'll only get alternative PRs.

ilovezfs:
Yeah, it's just sad they'd ship something that doesn't build and then wait a month but ho hum

DomT4:
I think it's another case of OS X/macOS not being a priority for an upstream, sadly.

ilovezfs:
Yeah, GNU gunk seems to have a serious quality control problem on macOS  to put it gently.

ilovezfs:
@DomT4 @MikeMcQuaid PR refreshed.

ilovezfs:
@DomT4 :green_apple:

ilovezfs:
`make check` parallelization(?) issue reported upstream: https://bugs.gnupg.org/gnupg/issue2425

ilovezfs:
New findings from https://github.com/Homebrew/homebrew-versions/pull/1322.

1) The "make check" failure in 2.1.14 is a regression from 2.1.13 (assuming the earlier patch which is now rolled into 2.1.14 has been applied as is done in the 2.1.13 version of the formula): http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1511/

2) The make check failure in 2.1.14 is not fixed in HEAD: http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1512/

3) There's a further regression in HEAD not found in 2.1.14.
```
Importing ssh keys... 
    > dsa key not added 
FAIL: ssh.scm
Checking passphrase cache (issue2015)... 
```

gnupg21.rb
```diff
@@ -17,12 +17,27 @@ class Gnupg21 < Formula
     depends_on "autoconf" => :build
     depends_on "automake" => :build
     depends_on "libtool" => :build
+
+    # prevents "/usr/local/opt/gettext/bin/msgfmt: found 866 fatal errors"
+    depends_on "gnu-sed" => :build
+
+    # provides `fig2dev`
+    depends_on "homebrew/science/transfig" => :build
```
DomT4: Given this formula will one day live (in one form or another) in `homebrew/core`, this line alone makes me think we should consider just dropping `HEAD` here.

DomT4:
I'd probably recommend filing `make check` failures to the [devel list](https://lists.gnupg.org/mailman/listinfo/gnupg-devel) rather than the upstream website. The former has a response time of less than a couple hours sometimes, compared to the year+ we've had stuff sitting on the website before now.

gnupg21.rb
ilovezfs: Nah, that's overkill. This can be avoided by passing `--disable-doc` as here: https://github.com/Homebrew/homebrew-versions/pull/1322/files#diff-c9a8a8a618586a787ca0fa9aea295653R71

DomT4: Well, it's not like the rest of `head` is giving you an easy ride here. It may be too unstable to merit bothering trying to support for now, but it's your PR to play with 😄.

ilovezfs: I haven't checked if any of those other deps drop away when `--disable-doc` is passed but I suspect they do, since this is just a discrepancy between building from the distributed tarball vs. building from the git repository, not any inherent instability at all.

ilovezfs:
@DomT4 the question is whether to ship with `make check` disabled since it fails that one test intermittently or toss out 2.1.14 until/unless there's a fix.

DomT4:
I'm uncomfortable ignoring `make check` failures in general, but particularly so for security-related software. I do think it's worth consulting the upstream mailing list on the issues you've unearthed & seeing if we can things resolved/patched.

ilovezfs:
Yeah, only reason I'd consider it here is that I've not been able to reproduce this failure locally, not even once.

DomT4:
And deparallelization didn't prove the answer to resolving the CI's woes there?

ilovezfs:
Nope, not at all.

ilovezfs:
LOL @DomT4 It's not a regression from 2.1.13 after all: http://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1537/version=yosemite/console

So looks like the patch that we had in there for 2.1.13 didn't fix the underlying issue after all. It's now integrated into 2.1.14, but equally ineffective, of course, since it's the same code.

ilovezfs:
Looks like this round was randomly a success. I'm inclined to ship since this issue is preexisting and not a regression.

ilovezfs:
@DomT4 upstream responded on the website: https://bugs.gnupg.org/gnupg/msg8660

>Thanks for letting us know.  Unfortunately, we do not test on MacOS yet, but we are working 
on that.

>I have neither experience with debugging on MacOS, nor do I have access to such a machine.  
I'm afraid you are on your own for now.

>The ssh test is new, so we need to figure out why it does not work.  Please do

>  make -C tests/openpgp check TESTS="setup.scm ssh.scm" verbose=2

>This lets us see what ssh-add prints to stderr.  It might be related to the version of 
OpenSSH shipped with the OS.


Also, I just provided additional findings here: https://bugs.gnupg.org/gnupg/msg8661, notably that the problem also occurs on 2.1.13, and that in two cases there were more specific error messages:
```
failed: gpgtar: gpg: [don't know]: invalid packet
(ctb=2d)
gpgtar: gpg: [don't know]: invalid packet (ctb=2a)
```
and
```
failed: gpgtar: gpg: Fatal: zlib inflate problem:
invalid block type
```

plttn:
FWIW, I'm not sure if tests cover this because it's a relatively niche feature, but at some point in bottle release history (currently tested on 2.1.13 compared to an unknown version on 2.1.xx), the `fetch` command to grab the public key from a smartcard URL appears to fail silently. It's possible this is just a weird interaction on my end (or an issue with gnupg itself) rather than a bottle issue, but I don't know enough to say for sure.

gnupg21.rb
```diff
@@ -1,28 +1,20 @@
...
-  head do
-    url "git://git.gnupg.org/gnupg.git"
-
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
-  end
-
   option "with-gpgsplit", "Additionally install the gpgsplit utility"
+  option "with-test", "Run `make check` prior to installing"
```
tdsmith: -1 on turning this off, sorry

ilovezfs: if you can make it fail locally, let me know.

tdsmith: if we want to ship this before understanding the failure i'd rather we did that by removing the bottle and keeping the tests mandatory

ilovezfs: I think that's a waste of electricity.

tdsmith: In a world with Bitcoin...

ilovezfs: got same stupid test to fail locally https://gist.github.com/ilovezfs/936c74c17b24f30c09598423e1cbbd63

all others passed

tdsmith:
~~ilz and I had a conversation about this on IRC and I think we can both live with "disable the component that we can't test"; the gpgtar test is the one with poor determinism and we can build the package --without-gpgtar so, yolo, let's ship.~~

other tests are failing too, and not just on the bot; the falcon cannot hear the falconer, the center cannot hold, etc etc

ilovezfs:
I cannot justify withholding the latest version of security software because of a new flaky test suite. The new testsuite uses tinyscheme, which, lest anyone forget, we boneyarded because it barely worked at all unless you built it with gcc.

ilovezfs:
Anyway, I'm going to close this issue and let others decide what to do about the flaky new test suite.

MikeMcQuaid:
If it's a security release I'm personally 👍 on shipping as-is, test suite be damned.

ilovezfs:
Yes, I think that does right by our users.

ilovezfs:
@BrewTestBot test this please.

tdsmith:
Is it a security release?

ilovezfs:
> * gpg: Properly ignore legacy keys in the keyring cache.

ilovezfs:
and from 2.1.14

> * gpgv: Change default options for extra security.

> * gpgsm: No more root certificates are installed by default.

ilovezfs:
Shipped. For anyone who's passionate about `make check` for this formula, PRs are welcome.

DomT4:
Given upstream don't have access to OS X/macOS, it's probably unfair to suggest `make check` is particularly supported by them and consequently expect it to pass, so I'm fine killing it for the time being. I suspect if we wanted to get it fixed we'll likely have to do some work ourselves.

ilovezfs:
Some good news on the `make check` problem. Upstream just got their hands on a Mac and reproduced the failure: https://bugs.gnupg.org/gnupg/msg9197

tdsmith:
:tada:

> Given upstream don't have access to OS X/macOS, it's probably unfair to suggest make check is particularly supported by them and consequently expect it to pass

Okay, but if they can't support their test suite, how can they support the rest of their software :|

MikeMcQuaid:
> Okay, but if they can't support their test suite, how can they support the rest of their software :|

Agreed.

Issue #1318, opened 2016-07-21
glfw3 3.2 fails to install on OS X 10.11


Marqin:
# Please follow the general troubleshooting steps first:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

_You can erase any parts of this template not applicable to your Issue._
### Bug reports:

I've tried to upgrade glfw3 via `brew upgrade` and it fails with

```
CMake Error at src/CMakeLists.txt:56 (set_target_properties):
set_target_properties called with incorrect number of arguments.
```

Full logs: https://gist.github.com/Marqin/9c8fa2982496071a16e6c1b258042675

ilovezfs:
What's the output of
```
brew outdated
```

Marqin:
`homebrew/versions/glfw3 (3.1.2 < 3.2)` and nothing more

2016-07-21 17:15 GMT+02:00 ilovezfs <notifications@github.com>:

ilovezfs:
And
```
cat /usr/local/Cellar/glfw3/3.1.2/INSTALL_RECEIPT.json
HOMEBREW_DEVELOPER=1 brew linkage glfw3
```

Marqin:
First:
{"used_options":["--without-shared-library"],"unused_options":["--universal","--with-examples","--with-tests"],"built_as_bottle":false,"poured_from_bottle":false,"time":1446984368,"HEAD":"73bb07d65dd5c56b88c51325992c50955e759e00","stdlib":null,"compiler":"clang","source":{"path":"/usr/local/Library/Taps/homebrew/homebrew-versions/glfw3.rb","tap":"homebrew/homebrew-versions","spec":"stable"}}

second - nothing

2016-07-21 17:23 GMT+02:00 ilovezfs <notifications@github.com>:

ilovezfs:
```
brew uninstall glfw3
brew install glfw3
```

Marqin:
That poured it from bottle, without errors.

2016-07-21 17:37 GMT+02:00 ilovezfs <notifications@github.com>:

ilovezfs:
@Marqin OK, do you mind trying
```
brew uninstall glfw3
brew install -dvs glfw3
```

Marqin:
Hmm, also no problems.

[...]

PR #1319, opened 2016-07-22
glfw3: fix static build

ilovezfs:
also add static test

### DIFF
--- a/glfw3.rb
+++ b/glfw3.rb
@@ -3,6 +3,7 @@ class Glfw3 < Formula
   homepage "http://www.glfw.org/"
   url "https://github.com/glfw/glfw/archive/3.2.tar.gz"
   sha256 "cb3aab46757981a39ae108e5207a1ecc4378e68949433a2b040ce2e17d8f6aa6"
+  head "https://github.com/glfw/glfw.git"
 
   bottle do
     cellar :any
@@ -25,9 +26,6 @@ class Glfw3 < Formula
 
   def install
     ENV.universal_binary if build.universal?
-
-    # make library name consistent
-    inreplace "CMakeLists.txt", /set\(GLFW_LIB_NAME\sglfw\)\n.*else\(\)\n/, ""
 
     args = std_cmake_args + %W[
       -DGLFW_USE_CHDIR=TRUE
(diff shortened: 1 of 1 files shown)


ilovezfs:
@DomT4 I'm not sure if there's still any need for the patching we were doing, but it caused https://github.com/Homebrew/homebrew-versions/issues/1318.

DomT4:
Old tree:
```
/usr/local/opt/glfw3
├── COPYING.txt
├── INSTALL_RECEIPT.json
├── README.md
├── include
│   └── GLFW
│       ├── glfw3.h
│       └── glfw3native.h
└── lib
    ├── cmake
    │   └── glfw3
    │       ├── glfw3Config.cmake
    │       ├── glfw3ConfigVersion.cmake
    │       ├── glfw3Targets-release.cmake
    │       └── glfw3Targets.cmake
    ├── libglfw3.3.2.dylib
    ├── libglfw3.3.dylib -> libglfw3.3.2.dylib
    ├── libglfw3.dylib -> libglfw3.3.dylib
    └── pkgconfig
        └── glfw3.pc

6 directories, 13 files
```

New tree:
```
~/Downloads/glfw3
└── 3.2
    ├── COPYING.txt
    ├── INSTALL_RECEIPT.json
    ├── README.md
    ├── include
    │   └── GLFW
    │       ├── glfw3.h
    │       └── glfw3native.h
    └── lib
        ├── cmake
        │   └── glfw3
        │       ├── glfw3Config.cmake
        │       ├── glfw3ConfigVersion.cmake
        │       ├── glfw3Targets-release.cmake
        │       └── glfw3Targets.cmake
        ├── libglfw.3.2.dylib
        ├── libglfw.3.dylib -> libglfw.3.2.dylib
        ├── libglfw.dylib -> libglfw.3.dylib
        └── pkgconfig
            └── glfw3.pc

7 directories, 13 files
```

Looks like this change effectively renames the library.

ilovezfs:
@DomT4 yep. upstream intentionally gives the static and dylib two different name. I have no idea why anyone was trying to mess with that.

DomT4:
Possibly worth bumping the revision if we're changing the primary `.dylib` versioning, and trying to build the `glbinding` formula with the `glfw3` option just to check it can still sniff it out alright.

This formula can actually go live in `homebrew/core` now really, but nobody's really put the effort into moving it across yet, including me.

apjanke:
Ping?

MikeMcQuaid:
@ilovezfs Thoughts here?

ilovezfs:
This formula should probably supplant the formula in core, and the upstream-undermining patch should go bye-bye.

MikeMcQuaid:
👍 from me.

ilovezfs:
Closing in favor of @JCount's superseding PR's:

https://github.com/Homebrew/homebrew-core/pull/7484
https://github.com/Homebrew/homebrew-versions/pull/1470

PR #1322, opened 2016-07-25
gnupg21: trigger CI

ilovezfs:
checking if "FAIL: gpgtar.scm" is reproducible with upstream HEAD
at 4ba11251aff578394000bf480f47160f0879c763.


apjanke:
Assuming this is a "do not merge", right?

ilovezfs:
I used this for bisecting a problem. Let's kill it.

PR #1323, opened 2016-07-25
Update cassandra22 dependencies

djchen:
Updates the Python cassandra-driver version to fix bug from https://issues.apache.org/jira/browse/CASSANDRA-11850

Also updated the other Python dependencies from the main/upstream Cassandra formula.


apjanke:
Merged. Thank you for your contribution to Homebrew!

PR #1329, opened 2016-08-02
boost160: new formula

tdsmith:


MikeMcQuaid:
Can you go for `boost16` as the name? Otherwise 👍

UniqMartin:
> Can you go for `boost16` as the name?

Sorry, this doesn't really make sense. The upstream version is 1.60.0, where the first component seems to be frozen forever and the middle component is incremented by one with every new Boost release. Rarely, there will be a .1 release if there are urgent bug fixes that can't wait for the next major Boost release.

DomT4:
This either needs a `conflicts_with` or `keg_only` status.

DomT4:
It may also be sane to put this in `homebrew/core`, and name it `boost+1.60.0` or something. That gives us a formula to play with IRT https://github.com/Homebrew/brew/issues/620. If we change our mind later it's easier to migrate seamlessly to taps than it is vice versa, for now.

tdsmith:
Added keg_only; I think `boost160` is the right name for `boost 1.60`.

@DomT4: can we prototype new version handling with something less urgent? :D

DomT4:
> @DomT4: can we prototype new version handling with something less urgent? :D

Yeah, not a problem at all. I can always tweak my OpenSSL 1.1.0 PR in the core to be the guinea pig; that's another huge breaking change coming up that'll definitely require its own formula for the foreseeable 😄.

PR #1330, opened 2016-08-02
boost155: generate 10.11 bottle

ilovezfs:


ilovezfs:
@DomT4 I'm assuming there's some reason this is *verboten*.

CC @MikeMcQuaid

DomT4:
Well, there's no need to force everyone to rebuild the whole formula to just update the bottle. Just stick a comma or a comment in somewhere & then drop this commit when you pull the updated bottle.

MikeMcQuaid:
> Just stick a comma or a comment in somewhere & then drop this commit when you pull the updated bottle.

Actually: just use the Homebrew Testing job on the bot.

DomT4:
Did we fix that in the end? Probably. It was problematic for a while after the core/brew repo split happened.

UniqMartin:
> Did we fix that in the end?

It should work again. If it doesn't, adding `--tap homebrew/versions` to the arguments for the Jenkins job will make sure none of the involved pieces get confused about the correct tap.

ilovezfs:
needs to be `--tap=homebrew/versions` but that seems to work. thanks @UniqMartin

ilovezfs:
Closed via https://github.com/Homebrew/homebrew-versions/commit/a01ff9d90bada17bb70fe35b51bd1c0f8cb5cbf5.

Surprisingly, I needed to use https://github.com/Homebrew/brew/pull/707 to actually pull the new bottles with
```
brew pull --bottle --tap=homebrew/versions https://bot.brew.sh/job/Homebrew%20Testing/1035/
```

PR #1337, opened 2016-08-10
lua53: fix "default-names" option for luarocks

telemachus:
Previously luarocks was installed as luarocks-5.3 and luarocks-admin-5.3
even if the user passed the `--with-default-names` switch.


lua53.rb
```diff
@@ -79,8 +79,13 @@ def install
...
         (share/"lua/5.3/luarocks").install_symlink Dir["#{libexec}/share/lua/5.3/luarocks/*"]
-        bin.install_symlink libexec/"bin/luarocks-5.3"
-        bin.install_symlink libexec/"bin/luarocks-admin-5.3"
+        if build.with? "default-names"
+          bin.install_symlink libexec/"bin/luarocks"
+          bin.install_symlink libexec/"bin/luarocks-admin"
+        else
+          bin.install_symlink libexec/"bin/luarocks-5.3"
+          bin.install_symlink libexec/"bin/luarocks-admin-5.3"
+        end
```
ilovezfs:
```
suffix = "-5.3" if build.without? "default-names"
bin.install_symlink libexec/"bin/luarocks#{suffix}"
bin.install_symlink libexec/"bin/luarocks-admin#{suffix}"
```

telemachus: Yup: that's neater. I did it that way and then amended the commit. (I had trouble when trying string interpolation earlier, but I must have done something else wrong. Thanks.)

ilovezfs: no problem ... thank you for opening this!

MikeMcQuaid:
Thanks again @telemachus!

PR #1344, opened 2016-08-19
move formulae to Formula/ dir

apjanke:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-versions/pulls) for the same formula update/change?
- [ ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)? _No. Formulae contents were not modified._
- [ ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)? _Not relevant_
- [X] Does your submission adhere to Versions "Acceptable Formulae" [guidelines](https://github.com/Homebrew/homebrew-versions/blob/07b17dfb9f1/README.md#acceptable-formulae)? _Doesn't change the formula set._

-----

Now that https://github.com/Homebrew/brew/issues/87 is fixed, we can move formula files around in repos without generating spurious `brew update` updated-formula messages.

This PR moves all the formula definition files in to a `Formula/` subdirectory. This keeps the tap tidy, separating the doco from the formula files, and makes the GitHub README page more readable (users don't have to scroll through a bunch of formulae to find the doco).

I tested this with a test copy of the repo, and updates done after the transition picked up the change without producing spurious messages.


DomT4:
Was this ever resolved?

> I think a more serious blocker is the FormulaVersions won't work after the formula directory change. As the result, we may get wrong revision number for bottles.

UniqMartin:
> Was this ever resolved?
>
> > I think a more serious blocker is the FormulaVersions won't work after the formula directory change. As the result, we may get wrong revision number for bottles.

No, but this was ignored when closing the referenced issue. Hence I'm currently :-1: on this.

apjanke:
Makes sense. Aborting this for now until FormulaVersions is fixed.

MikeMcQuaid:
Git can/does track renames so this _should_ be relatively straightforward to do.

PR #1349, opened 2016-08-20
node4-lts: Permit weak imports

alanorth:

-----

Permit weak imports. Fixes build on macOS 10.12 with Xcode 8.0 >= Beta 4.


MikeMcQuaid:
Any idea what the weak imports are and why CC @DomT4.

alanorth:
Here is the error:

```
ld: weak import of symbol '_posix_memalign' not supported because of option: -no_weak_imports for architecture x86_64
```

And here is `brew gist-logs node4-lts` after the failure: https://gist.github.com/12c6f281a60e1872251a78ab93fe6fec

ilovezfs:
My guess is it's because it's hard coding the deployment target.
@alanorth do you mind trying
```
inreplace "common.gypi", "'MACOSX_DEPLOYMENT_TARGET': '10.5'", "'MACOSX_DEPLOYMENT_TARGET': '#{ENV["MACOSX_DEPLOYMENT_TARGET"]}'"
```

*without* your patch in the PR.

alanorth:
Sure, I can try that, @ilovezfs. I don't where to put that code, though. Thanks.

ilovezfs:
same place where you have your permit weak imports! first line of install.

DomT4:
I've a feeling the deployment target here is legit in the same way Go's is.

Node bumped the `MACOSX_DEPLOYMENT_TARGET` to 10.7 for the current stable, which doesn't seem to have this problem _(unless I managed to compile `node` in my 10.12 box before we merged the weak link change & am misremembering)_ & I guess we could mimic if necessary.

ilovezfs:
Whether legit or not, we need to know if that's the cause.

DomT4:
Agreed, I was just elaborating that I believe this is justified behaviour as far as upstream are concerned. The failure isn't actually coming from Node themselves but rather `v8`.

ilovezfs:
@DomT4 yeah, I think this disallow weak imports thing is not going to play nicely with software that wants to deliberately set a lower deployment target. I'm pretty sure this should be opt in not opt out :)

DomT4:
IMO, the general rule should be something like _"If it can be used as a compiler, weak links should be allowed"_.

ilovezfs:
I think the rule should be: Apple made it opt in not opt out, so let's only set when we have to.

DomT4:
```
==> Summary
🍺  /usr/local/Cellar/node/6.4.0: 3,829 files, 39.8M, built in 9 minutes 42 seconds
```
Definitely not a problem on 10.12 with the current Node with the 10.7 deployment target & likely newer bundled `v8` for whatever reason.

alanorth:
@ilovezfs it still fails with (only) your patch, but now with a different error:

```
==> make install
Last 15 lines from /Users/aorth/Library/Logs/Homebrew/node4-lts/02.make:
In file included from ../src/async-wrap-inl.h:6:
In file included from ../src/base-object-inl.h:5:
In file included from ../src/env.h:5:
In file included from ../src/debug-agent.h:25:
../src/util.h:12:10: fatal error: 'tr1/type_traits' file not found
#include <tr1/type_traits>  // NOLINT(build/c++tr1)
         ^
1 error generated.
make[1]: *** [/private/tmp/node4-lts-20160821-30445-1r9jij7/node-v4.5.0/out/Release/obj.target/node/src/debug-agent.o] Error 1
make[1]: *** Waiting for unfinished jobs....
1 error generated.
make[1]: *** [/private/tmp/node4-lts-20160821-30445-1r9jij7/node-v4.5.0/out/Release/obj.target/node/src/async-wrap.o] Error 1
1 error generated.
make[1]: *** [/private/tmp/node4-lts-20160821-30445-1r9jij7/node-v4.5.0/out/Release/obj.target/cctest/test/cctest/util.o] Error 1
make: *** [node] Error 2
```

ilovezfs:
Is it before or after the old failure location?

ilovezfs:
@alanorth btw you can also try
```
inreplace "common.gypi", "'MACOSX_DEPLOYMENT_TARGET': '10.5'", "'MACOSX_DEPLOYMENT_TARGET': '10.6'"
```

```

inreplace "common.gypi", "'MACOSX_DEPLOYMENT_TARGET': '10.5'", "'MACOSX_DEPLOYMENT_TARGET': '10.7'"
```
etc.

and see if any of those work.

ilovezfs:
```
int      posix_memalign(void **, size_t, size_t) __OSX_AVAILABLE_STARTING(__MAC_10_6, __IPHONE_3_0);
```

So looks like 10.6 would be the one to try!

DomT4:
I could live with bumping to 10.6/10.7 on the deployment target if necessary, and if it works. That would echo the current stable branch of Node.

ilovezfs:
I'd like to know if it builds with 10.6 or 10.7 with the no weak imports thing, but I think we might as well go with the weak imports enabled until other people are on board with only making it opt-in. Since the symbol wasn't defined until 10.6 and upstream sets a deployment target of 10.5, we're just creating a bug that doesn't actually exist. So let's just not do that.

alanorth:
Ok, it builds with target of 10.6.

DomT4:
> until other people are on board with only making it opt-in

I don't think we can realistically make it opt-in. Theoretically there are thousands and thousands of formulae this sort of thing is absolutely fine with, with the exception of things that deliberately set a deployment target way back for "lazy" backwards compatibility when using that tool to compile other projects.

There are real problems with not using the flag, and evidently... real problems with using the flag as well, lol. We're kind of rock & a hard place on this.

ilovezfs:
>I don't think we can realistically make it opt-in.

Most upstream projects are not going to be compiling with that flag. Of course we can make it just opt-in.

alanorth:
Both Mac OS X 10.6 and the original Node.js were both released in 2009. Prohibiting weak imports is supposed to make software safer, so maybe we can just be OK with saying that, in Homebrew, "_node v4 LTS runs on Mac OS X 10.6 and up_"?

MikeMcQuaid:
This will be handled in https://github.com/Homebrew/brew/pull/782 instead.

PR #1358, opened 2016-08-31
gsl1 1.16 (new formula) and gsl114: remove

ilovezfs:
On the advice of @tschoonj:

>I would replace 114 with 116 as the latter is fully backwards compatible.

See https://github.com/Homebrew/homebrew-core/pull/4341#issuecomment-243666378


MikeMcQuaid:
Maybe worth just making this `gsl1`?

gsl116.rb
```diff
@@ -0,0 +1,21 @@
+class Gsl < Formula
```
MikeMcQuaid: Class name needs fixed: https://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1596/version=el_capitan/console

ilovezfs:
Yeah, we have precedent for `foo1`:
```
jboss-forge1.rb
libcouchbase1.rb
raptor1.rb
```
so that sounds fine.

PR #1360, opened 2016-08-31
logstash 2.3.4

jasontedor:
This commit removes the logstash22 formula as the Logstash 2.2 series is
now unmaintained, and adds an logstash23 formula for the 2.3 series of
Logstash.


MikeMcQuaid:
Thanks again @jasontedor!

jasontedor:
Right back at you @MikeMcQuaid! :smile:

PR #1366, opened 2016-09-04
Solr 5.4 brew recipe

damienmckenna:
This includes a recipe for Solr 5.4.1 named "solr54".


MikeMcQuaid:
An audit failure: https://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1605/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_versions_solr54___new_formula/

Also, can you delete `solr4.rb`as part of this PR? Thanks!

damienmckenna:
Is there any reason to drop Solr4? Are there any stats on its usage?

damienmckenna:
BTW I just copied the solr55.rb file to make the solr54.rb so the error exists in it too. Maybe all files should be updated to put the depends_on lines before the conflicts_with lines?

MikeMcQuaid:
> BTW I just copied the solr55.rb file to make the solr54.rb so the error exists in it too. Maybe all files should be updated to put the depends_on lines before the conflicts_with lines?

If you can update both it'd be ❤️ 

> Is there any reason to drop Solr4? Are there any stats on its usage?

https://github.com/Homebrew/homebrew-versions#acceptable-formulae basically.

damienmckenna:
I was doing this PR because the internal logic in Solr changed between 5.4 and 5.5 related to AND queries, so one of the Drupal 7 integration systems (Search API Solr) would give incorrect results. It may just be better to leave the Solr formulae as is and we'll just fix the Drupal 7 thing?

FYI I moved the fix-the-existing-formulae thing into a separate PR: #1368

MikeMcQuaid:
@damienmckenna Happy to add this just want to avoid having more than those 2 versions in a single PR so if you could remove it here that'd be great otherwise I'll add to me list to do so later before we merge this.

solr54.rb
```diff
@@ -0,0 +1,69 @@
...
+  desc "Enterprise search platform from the Apache Lucene project"
+  homepage "https://lucene.apache.org/solr/"
+  url "https://www.apache.org/dyn/closer.cgi?path=lucene/solr/5.4.1/solr-5.4.1.tgz"
+  mirror "https://archive.apache.org/dist/lucene/solr/5.4.1/solr-5.4.1.tgz"
+  sha256 "3e4b4ec7bd728b49b2ebc3dbe8f3d1ef89fded4ab86b9e2f856bedd58c99f28b"
+
+  depends_on :java
+
+  conflicts_with "solr", :because => "Differing versions of the same formula"
+  conflicts_with "solr4", :because => "Differing versions of the same formula"
```
MikeMcQuaid: This line needs removed now the formula is gone. Then we should be 👍 to 🚢

damienmckenna:
I updated the conflicts_with lines for both the solr54 and solr55 formulae.

MikeMcQuaid:
`audit` is failing and conflicts need resolved: https://bot.brew.sh/job/Homebrew%20Versions%20Pull%20Requests/1626/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_versions_solr54___new_formula/

damienmckenna:
I've rebuilt this PR in a way that'll make it easier to fix merge conflicts: #1382

PR #1377, opened 2016-09-05
Add Liquigraph recipe (v2.0.0)

fbiville:

---

I have two parallel versions of Liquigraph (feature-wise, they're equivalent):
- Liquigraph 2.0.0 works with Neo4j 2.x (this PR)
- Liquigraph 3.0.0 works with Neo4j 3.x ([PR against main repo](https://github.com/Homebrew/homebrew-core/pull/509))

I need to have both versions available to users so they can pick the one that suits their setup. It seems I have to pick a default, that's why I went with `3.0.0` in the main repo.

### DIFF
--- a/liquigraph200.rb
+++ b/liquigraph200.rb
@@ -1,0 +1,38 @@
+class Liquigraph200 < Formula
+  desc "Migration runner for Neo4j"
+  homepage "http://www.liquigraph.org"
+  url "https://github.com/fbiville/liquigraph/archive/liquigraph-2.0.0.tar.gz"
+  sha256 "2c9232aa4db0d8724733643b6a919910899d8e9d49965bf48dbda044c9a00021"
+  head "https://github.com/fbiville/liquigraph.git"
+
+  depends_on "maven" => :build
+  depends_on :java => "1.7+"
+
+  def install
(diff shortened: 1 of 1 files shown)


fbiville:
Hopefully, the notability constraint can be ignored here as well? :angel: https://github.com/Homebrew/homebrew-core/pull/509#issuecomment-213300733

MikeMcQuaid:
@fbiville In this case it's more that we don't package Neo4J 2.0 so I doubt this package would be used?

fbiville:
@MikeMcQuaid Neo4j 2.x is still deployed in production quite a lot, so I think many devs still have a local installation of Neo4j v2 (maybe not via homebrew), that's why I think it could be valuable to have Liquigraph 2.0.0.

I can keep it as a tap, anyway.

MikeMcQuaid:
Another option: I'd be prepared to accept this if you add neo4j 2 as part of this PR too 😀

MikeMcQuaid:

fbiville:
Yep sorry, no time to work on this.
And I've been told there is already a Neo4j v2 homebrew recipe, so I'm a bit confused :-)

PR #1382, opened 2016-09-08
Replace the Solr 4 formula with Solr 5.4.

damienmckenna:
This makes several changes:
* The Solr 4 formula has been removed (per request by MikeMcQuaid).
* A new formula for Solr 5.4(.1) has been added with the name "solr54".
* The existing Solr 5.5 formula has been updated to fix to move the
  conflicts_with lines before the depends_on line, and to indicate it
  conflicts with the new 5.4 formula.

(it also replaces PR #1366 which was built wrong)


apjanke:
Test failure is just an audit failure.

```
Error: 1 problem in 1 formula
homebrew/versions/solr54:
  * `bottle modifier` (line 13) should be put before `depends_on` (line 8)
```

Could you fix that up?

damienmckenna:
I've fixed that issue and the 'conflicts_with' problem on the Solr55 formula.

MikeMcQuaid:
Thanks again @damienmckenna!

PR #1384, opened 2016-09-14
LLVM 3.9 formula

SASinestro:

---

It's just a tweak of the not-yet-merged formula for LLVM 3.9 from `homebrew-core`.

### DIFF
--- a/llvm39.rb
+++ b/llvm39.rb
@@ -1,0 +1,400 @@
+class CodesignRequirement < Requirement
+  include FileUtils
+  fatal true
+
+  satisfy(:build_env => false) do
+    mktemp do
+      cp "/usr/bin/false", "llvm_check"
+      quiet_system "/usr/bin/codesign", "-f", "-s", "lldb_codesign", "--dryrun", "llvm_check"
+    end
+  end
+
+  def message
+    <<-EOS.undent
+      lldb_codesign identity must be available to build with LLDB.
+      See: https://llvm.org/svn/llvm-project/lldb/trunk/docs/code-signing.txt
+    EOS
+  end
+end
+
+class Llvm39 < Formula
+  desc "Next-gen compiler infrastructure"
+  homepage "http://llvm.org/"
+
+  stable do
+    url "http://llvm.org/releases/3.9.0/llvm-3.9.0.src.tar.xz"
+    sha256 "66c73179da42cee1386371641241f79ded250e117a79f571bbd69e56daa48948"
+
+    resource "clang" do
+      url "http://llvm.org/releases/3.9.0/cfe-3.9.0.src.tar.xz"
+      sha256 "7596a7c7d9376d0c89e60028fe1ceb4d3e535e8ea8b89e0eb094e0dcb3183d28"
+    end
+
+    resource "clang-extra-tools" do
+      url "http://llvm.org/releases/3.9.0/clang-tools-extra-3.9.0.src.tar.xz"
+      sha256 "5b7aec46ec8e999ec683c87ad744082e1133781ee4b01905b4bdae5d20785f14"
+    end
+
+    resource "compiler-rt" do
+      url "http://llvm.org/releases/3.9.0/compiler-rt-3.9.0.src.tar.xz"
+      sha256 "e0e5224fcd5740b61e416c549dd3dcda92f10c524216c1edb5e979e42078a59a"
+    end
+
+    # Only required to build & run Compiler-RT tests on OS X, optional otherwise.
+    # http://clang.llvm.org/get_started.html
+    resource "libcxx" do
+      url "http://llvm.org/releases/3.9.0/libcxx-3.9.0.src.tar.xz"
+      sha256 "d0b38d51365c6322f5666a2a8105785f2e114430858de4c25a86b49f227f5b06"
+    end
+
+    resource "libunwind" do
+      url "http://llvm.org/releases/3.9.0/libunwind-3.9.0.src.tar.xz"
+      sha256 "66675ddec5ba0d36689757da6008cb2596ee1a9067f4f598d89ce5a3b43f4c2b"
+    end
+
+    resource "lld" do
+      url "http://llvm.org/releases/3.9.0/lld-3.9.0.src.tar.xz"
+      sha256 "986e8150ec5f457469a20666628bf634a5ca992a53e157f3b69dbc35056b32d9"
+    end
+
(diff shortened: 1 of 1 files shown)


SASinestro:
I apparently can't manage to use git. Joy.

SASinestro:
Wait, I just realized that it doesn't actually work, and that the symlinks that it creates aren't actually getting the version appended to them like they're supposed to.

I can't see why immediately.

MikeMcQuaid:
> Wait, I just realized that it doesn't actually work, and that the symlinks that it creates aren't actually getting the version appended to them like they're supposed to.

Can you elaborate on "doesn't actually work" if you'd like help? 

To help us debug this issue can you explain:
- What you were trying to do (and why)
- What happened
- What you expected to happen
- Step-by-step reproduction instructions (with as minimal input data as possible)

Thanks!

SASinestro:
Sorry @MikeMcQuaid, I figured that it was pretty unambiguous but I can see that I wasn't very clear.

The issue is that the code I copied from the `llvm38` formula to symlink the llvm binaries into `/usr/local/bin` doesn't actually change the names like it does in that formula, so you end up with just `llvm-config` et al. instead of the proper `llvm-config-3.9` (and so on) in the path, which has obvious problems.

Just installing it is enough to replicate it.

MikeMcQuaid:
@SASinestro Ok, thanks for the explanation.

Godin:
@SASinestro wield guess - maybe what is missing is
```
install_prefix = lib/"llvm-#{ver}"
```
? See https://github.com/Homebrew/homebrew-versions/blob/19e6a94df3130a8038789aa5c1dcd5eccffc5850/llvm38.rb#L173

In #1395 I tried to be fully based on `homebrew-versions/llvm38` instead of unfinished update of `homebrew-core/llvm` ( https://github.com/Homebrew/homebrew-core/pull/4579 ) and not the other way around. Closed it in favor of your changes since you was first one to submit and as anyway it wasn't passing build. Let me know, if you'd need help with finalization of this - I might give it a new try.

And BTW at the end this supposed to close #1388 .

Marqin:
@Godin I think you shouldn't close your PR if this one is not working and yours is working.

Godin:
@Marqin I didn't said that mine was working, I said exactly opposite - it was also requiring tuning.

georgemarrows:
In a gentle attempt to move this along a little, I've created a Gist outlining my current attempts to create a llvm39.rb formula. It builds, but not much else. See https://gist.github.com/georgemarrows/a4fb838114eae0504abe3fd6d8fd5cc3
Please let me know if the Homebrew community doesn't take kindly to such actions (I see Godin deleted his duplicate pull request). 

As noted in the Gist, this takes a different tack from Godin: I deliberately chose to make llvm39.rb look like the current homebrew-core/llvm formula. The (obvious) pros and cons are listed in the gist.

MikeMcQuaid:
The best thing to do now would be to create a `keg_only` `llvm@3.9` formula in homebrew/core.

georgemarrows:
@MikeMcQuaid I'm not very knowledgeable on Homebrew, but previous fixed versions of LLVM all seem to have been eg `llvm38` in homebrew-versions. Can I just check if you're proposing something different for 3.9? If so, can you point me at docs or an example of an xxx@version formula in homebrew/core? I'm not sure how to specify an `llvm@3.9` formula.

Btw sorry for the v slow reply - been travelling. Should be able to get back to this from Monday, but am very happy for others to take over.

ilovezfs:
Can we not go straight to 3.9 assuming :green_apple: for CI?

MikeMcQuaid:
Yeh, we can in this case. A few weeks ago or for the next major version we may want to investigate the versioned approach again but for now we're fine as-is.

ilovezfs:
This is not a good candidate for versioning IMO, but that can be discussed at a later date.

Godin:
@MikeMcQuaid Would you mind if I'll open PR with `llvm39.rb` , which is [homebrew-core/Formula/llvm.rb](https://github.com/Homebrew/homebrew-core/blob/a3bd7f57a4432b17274964b3b3599a56453f6dd7/Formula/llvm.rb) with a following patch for installation into separate directory? 
```diff
20c20
< class Llvm < Formula
---
> class Llvm39 < Formula
169a170,174
>   # version suffix
>   def ver
>     "3.9"
>   end
>
212a218
>     install_prefix = lib/"llvm-#{ver}"
215a222
>     args << "-DCMAKE_INSTALL_PREFIX=#{install_prefix}"
270c277
<     inreplace "#{share}/clang/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", "#{bin}/clang"
---
>     inreplace "#{share}/clang/tools/scan-build/bin/scan-build", "$RealBin/bin/clang", "#{bin}/clang-#{ver}"
275,276c282,293
<     (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm"
<     (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang"
---
>     (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm" => "llvm-#{ver}"
>     (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang" => "clang-#{ver}"
>
>     Dir.glob(install_prefix/"bin/*") do |exec_path|
>       basename = File.basename(exec_path)
>       bin.install_symlink exec_path => "#{basename}-#{ver}"
>     end
>
>     Dir.glob(install_prefix/"share/man/man1/*") do |manpage|
>       basename = File.basename(manpage, ".1")
>       man1.install_symlink manpage => "#{basename}-#{ver}.1"
>     end
```

My basic manual tests pass, but not excluded that it requires polishing and I forgot something that BrewBot will catch. I feel like should ask, given that other threads on same subject were locked. Whatever the decision is - fine by me, I can live without versioned formula, just thought that can try to help Homebrew users on this.

MikeMcQuaid:
@Godin We've historically not accepted formulae into homebrew/versions that match the current stable version. I'd probably rather than continued but have no strong feelings. Is this something you personally need? If it's just for a small number of people we'd rather leave it as-is. They can add it to their own tap if they need to pin it to a specific version.

Godin:
@MikeMcQuaid ah, okay, didn't know about this. And not a problem for me. This is simply in reaction to all the buzz around subject that I've got in inbox being an initiator of #4579. Thanks.

MikeMcQuaid:

PR #1395, opened 2016-09-23
llvm 3.9.0

### LINKED ISSUE #1388: add llvm39
LLVM 3.9 Was realesed. 

It's [problematic](https://github.com/Homebrew/homebrew-core/pull/4579) to update normal `llvm` package to 3.9, but it should be easy to just deploy it as `llvm39` package.

Godin:
Based on existing `llvm38`.

This closes #1388.


Godin:
Oops, actually duplicate of #1384

Marqin:
I think you should reopen this, because #1384 is not maintained.

MikeMcQuaid:
@Marqin Feel free to open a new PR.

Marqin:
@MikeMcQuaid I'm not proficient enough in LLVM to know how to configure new version properly. And given LLVM is big project, than much could have changed.

Godin:
@Marqin Excuse me, but then could you please stop dictating what others should do - this is not really helping. You need to patiently wait while others do their best in configuring new version properly, exactly because this is not that easy and "LLVM is big project".

Marqin:
@Godin no. I'll say what I think is good for the project.

DomT4:
There's decent process ongoing in https://github.com/Homebrew/homebrew-core/pull/4579.

@MikeMcQuaid If we want to add a versioned formulae I'd be tempted by `llvm@3.9` in core at this point. There's enough support in place to start using that system a teeny bit more broadly IMO.

MikeMcQuaid:
> @MikeMcQuaid If we want to add a versioned formulae I'd be tempted by llvm@3.9 in core at this point. There's enough support in place to start using that system a teeny bit more broadly IMO.

@DomT4 Yes, please 😁

PR #1338, opened 2016-08-15
Add bazel 0.2.3

petemounce:
... because we need to peg the version so we can automate building CI workers while we update to support the later 0.3.1. We're happy to maintain past versions of bazel, here, since we're standardising on it as our build tool.

- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-versions/pulls) for the same formula update/change?
  - I found #1306 (closed) which suggests that you might not want to accept this PR, but I wasn't sure - apologies if this is unwelcome in the repo.
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?
- [x] Does your submission adhere to Versions "Acceptable Formulae" [guidelines](https://github.com/Homebrew/homebrew-versions/blob/07b17dfb9f1/README.md#acceptable-formulae)?
  - I think so? Of the ones I think apply:
  - [x] am within +/- 2 major/minor versions
  - [x] conflicts_with bazel

PR #1415, opened 2016-10-11
Add Bazel 0.3.1 for version pegging

petemounce:
- [ ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?
  - In #1338 I faithfully promised to maintain these, so, hi again :)

-----

Please note: https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md now 404s.


MikeMcQuaid:
This should be `Bazel03`. We won't accept old patch releases. If you need them please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

ilovezfs:
@petemounce are you sure this is needed. I'm surprised we've not heard from others who use Bazel asking for this if there are significant compatibility issues here.

petemounce:
@MikeMcQuaid I don't follow - [bazel02.rb](https://github.com/Homebrew/homebrew-versions/blob/master/bazel02.rb) pulls bazel 0.2.3 (latest 0.2.x), not 0.2.0. I'm following that pattern, on that precedent.

@ilovezfs bazel 0.3.2 was released on Friday 7 October, so it seems reasonable for there to not have been users asking for this yet - only two working days since the release. I use bazel at work, and I'm currently updating us (and our build boxes) from 0.3.1 to 0.3.2, but our configuration management (puppet in this case) only has the option of `brew install bazel` to get latest, as opposed to a particular version, without this change (at least, as I understand how homebew works, which is a shallow-at-best view so far).

ilovezfs:
You can give a URL to brew install.

MikeMcQuaid:
Homebrew does not wish to support "pin to security/patch version". This is best achieved by maintaining your own tap. Sorry!

petemounce:
@MikeMcQuaid that sounds like "why not?" might be a common FAQ - would you mind linking me to the reason? Closing this PR seems like it's inconsistent with the existence in this repo of bazel02.rb. Bazel is quite new software, and very actively developed, so at the moment even patch releases are necessary to peg.

Is there some documentation on how to set up our own tap _with_ the automation that your bot has for building bottles, so that we can benefit from that too, please?

MikeMcQuaid:
@petemounce I'm hoping in the medium-term to 🔥 this tap entirely so I'm trying to spend the limited time on working towards that goal rather than better supporting it. Point taken, though. Similarly on making it easier to set up your own taps to do this: it's on my list.

petemounce:
@MikeMcQuaid is there anything that is low hanging fruit that I could help with? I ruby a bit. I'm also happy to try out unstable builds if that would help.

re: my own tap - @jordigg (colleague) pointed out we can just manually pull in the iterations of homebrew-core/Formula/bazel.rb as we bless them internally, and just adjust the formula name so we don't collide with the one in homebrew-core and prefer that one. We'll therefore get the bottles that your bot produces.

MikeMcQuaid:
> re: my own tap - @jordigg (colleague) pointed out we can just manually pull in the iterations of homebrew-core/Formula/bazel.rb as we bless them internally, and just adjust the formula name so we don't collide with the one in homebrew-core and prefer that one. We'll therefore get the bottles that your bot produces.

If you change the name you won't get our bottles 😢. However, if you install the fully-qualified-name (e.g. `brew install yourorg/yourtap/bazel`) then I think you could/will.

> @MikeMcQuaid is there anything that is low hanging fruit that I could help with? I ruby a bit. I'm also happy to try out unstable builds if that would help.

Thanks for the offer! Re: docs: feel free to open PRs to amend things as to your best understanding and I'll offer suggestions and help. A secret: there's so much on my todo list the easiest thing to do is to do 1-10% of the work and then ask me for help on the remaining 90% and I probably will 😉

petemounce:
Is there an existing GH issue relating to the changes you intend to happen that I can read?

I mean, as a totally naive never-read-the-code-before stab, it sounds like `Formula` could be adjusted so that it maybe takes a `versions` array, and objects within that with `url`, `sha256`, `bottle` attributes, and perhaps also `install` and `test` lambdas (or make it possible to tag the existing DSL methods to peg them to particular versions, if they tend to change between versions in the first place).

Then the thing that uses `Formula` would ideally get a compatibility layer so that it would be possible to treat all the current `Formula` instances as a new-style Formula with a single version, so that the change doesn't break the world...?

MikeMcQuaid:
https://github.com/Homebrew/brew/issues/620 is the best place to read.

jordigg:
@MikeMcQuaid is there any way we can just clone this repo or just create our own one but still use the already compiled bottles from homebrew-core hosted on Bintray https://bintray.com/homebrew/bottles/bazel?

MikeMcQuaid:
@jordigg As said above: create your own tap and copy files that way or alternatively stop your users from running `brew update`.

jordigg:
We already did that but when looks for the bottle it fails and builds from source because can't find it. I was wondering if there is an easy way to point our copy of the bazel.rb to the "core" cellar.
Maybe with something similar to
```
bottle do
  root_url "https://bintray.com/homebrew/"
  ...
```

The problem isn't an update, is a new machine being set up (build server) which gets the latest version from brew and that specific version brakes our CI. I can install using the bazel.sh file from the official repo but wanted to find a solution on pinning versions on our own repo using brew.
Thanks for your help!

MikeMcQuaid:
@jordigg Yeh, something like that should work.

MikeMcQuaid:
@petemounce https://github.com/Homebrew/homebrew-versions/commit/cb3ca696b6a35991953706c355dcb262684ec5bc?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

MikeMcQuaid:
@petemounce Updated the README based on our chat: https://github.com/Homebrew/homebrew-versions/commit/cb3ca696b6a35991953706c355dcb262684ec5bc?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

petemounce:
Thanks, that makes it clearer.

By the way, the PR template in here links to a contributing.md that used to exist but now does not.

MikeMcQuaid:
@petemounce Fixed, cheers.

PR #1416, opened 2016-10-11
apache-spark20 2.0.1 (new formula)

mileserickson:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/.github/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-versions/pulls) for the same formula update/change?
- [X] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [X] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?
- [X] Does your submission adhere to Versions "Acceptable Formulae" [guidelines](https://github.com/Homebrew/homebrew-versions/blob/07b17dfb9f1/README.md#acceptable-formulae)?

---

### DIFF
--- a/apache-spark20.rb
+++ b/apache-spark20.rb
@@ -1,0 +1,25 @@
+class ApacheSpark20 < Formula
+  desc "Engine for large-scale data processing"
+  homepage "https://spark.apache.org/"
+  url "https://www.apache.org/dyn/closer.lua?path=spark/spark-2.0.1/spark-2.0.1-bin-hadoop2.7.tgz"
+  version "2.0.1"
+  sha256 "3d017807650f41377118a736e2f2298cd0146a593e7243a28c2ed72a88b9a043"
+  head "https://github.com/apache/spark.git"
+
+  bottle :unneeded
+
+  conflicts_with "apache-spark", :because => "Differing version of same formula"
+
+  def install
+    # Rename beeline to distinguish it from hive's beeline
+    mv "bin/beeline", "bin/spark-beeline"
+
+    rm_f Dir["bin/*.cmd"]
+    libexec.install Dir["*"]
+    bin.write_exec_script Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    system "#{bin}/spark-shell <<<'sc.parallelize(1 to 1000).count()'"
+  end
+end


mileserickson:
Submitting to freeze Apache Spark version at 2.0 in an installation script.

MikeMcQuaid:
This is the current version in homebrew/core so want to get more thoughts before including this.

> Submitting to freeze Apache Spark version at 2.0 in an installation script.

A better fit for freezing dependencies is to create your own tap.

mileserickson:
Thanks, Mike!

This is my first foray into `homebrew-versions`, and I certainly don't have standing to rock the boat when this system clearly works well for an awful lot of people. If you'd prefer that this be tapped privately, that's OK.

Original rationale: my expectation is that the eventual release of Spark 2.1.0 is highly likely to introduce new and unpredictable bugs, especially on OSX. The motivation is that I do not want to instruct my audience (students, in this case) to `brew install apache-spark` without specifying a major+minor version number that is verified to work. It's cheap insurance against future firefighting.

Separately, I'm curious about the engineering rationale behind waiting until obsolescence to add items to `homebrew-versions`; as a systems engineer, I'm inclined to think that automatically adding "new" versions to `homebrew-versions` as soon as they are added to `homebrew-core` could reduce the reactive workload of adding "old" versions later. (After all, everything will be "old" soon enough!) But, that's almost certainly a conversation for another forum.

Regardless, I'm deeply grateful for everything that you do here. Thanks again.

MikeMcQuaid:
> Original rationale: my expectation is that the eventual release of Spark 2.1.0 is highly likely to introduce new and unpredictable bugs, especially on OSX.

The best way you can help and/or avoid that is to help add a more comprehensive test to the `apache-spark` formula in homebrew/core so we can catch issues like that before release and/or to consider helping upstream test their software before it's released.

> The motivation is that I do not want to instruct my audience (students, in this case) to brew install apache-spark without specifying a major+minor version number that is verified to work. It's cheap insurance against future firefighting.

Sure, I get that. In this case the best option may be to use a tap instead then you have absolute control. Technically, 2.1.0 could be all ✨ and 🌈 and 2.1.1 could be awful and then we'd not protect you from that. I realise this is more work for you but the current approach is more work for us.

> Separately, I'm curious about the engineering rationale behind waiting until obsolescence to add items to homebrew-versions; as a systems engineer, I'm inclined to think that automatically adding "new" versions to homebrew-versions as soon as they are added to homebrew-core could reduce the reactive workload of adding "old" versions later. (After all, everything will be "old" soon enough!) But, that's almost certainly a conversation for another forum.

Automatically adding them would massively balloon the number of packages included in this repo and thus the ones we have to support and deal with issues being files on them.

If we could do Homebrew again from scratch today and have no legacy codebase and a experienced, full-time paid team we'd probably have every major/minor version kept, packaged and with binaries available for all time. Unfortunately we are extremely resource constrained so need to take a more balanced approach which means keeping this tap and our versions approach in general more focused on the packages where it causes us (rather than just end-users) pain.

Thanks for your kind words and lovely tone.

mileserickson:
Spark 2.1 has been released. Is this pull request now appropriate?

mileserickson:
I'm noticing that homebrew-versions is now identified as deprecated and that versioned formulae are migrating into homebrew-core.

This looks like fantastic news and I'd love to help, particularly if this introduces a mechanism for freezing the current version of a project.

However, I'm struggling to find the announcement. I checked the Homebrew Blog, searched both discourse.brew.sh and Google, and reviewed the homebrew-versions README here on GitHub, but I'm drawing blanks.

Is there a way to help with the implementation or documentation of this change?

Cheers,

Miles

MikeMcQuaid:
> Is there a way to help with the implementation or documentation of this change?

@mileserickson Apologies that the documentation was lacking; I forgot to commit the changes to the README of this repository. I've done so in https://github.com/Homebrew/homebrew-versions/commit/c94a1b18d5c2c198b4be302b604dd0c845a4344b which links to the main documentation in http://docs.brew.sh/Versions.html. The relevant issue is in https://github.com/Homebrew/brew/issues/620 which tracks some of the longer-term aspirations and goals.

> This looks like fantastic news and I'd love to help, particularly if this introduces a mechanism for freezing the current version of a project.

The current goal is just to migrate everything here into homebrew/core that can be. I'd love your help with this. The best way to do so would be to create PRs like https://github.com/Homebrew/homebrew-core/pull/8376 where you migrate formulae over to the new format.

Freezing the current version is not offered yet but is a longer-term goal (but note we'll only allow freezing on major/minor and not patch/security releases).

> However, I'm struggling to find the announcement. I checked the Homebrew Blog, searched both discourse.brew.sh and Google, and reviewed the homebrew-versions README here on GitHub, but I'm drawing blanks.

There has been no announcement yet. I'm going to wait until at least this tap has been official migrated over and see if I can't get some of the another niceties working and then ship Homebrew 1.2.0 with this work.

mileserickson:
Outstanding, thanks. Let's do this.

PR #1418, opened 2016-10-12
allow use with gpgme

jeroenooms:
The current version of gpgme 1.7 has a bunch of new features specifically to work with GnuPG 2.1. See for example the [new keygen functions](https://www.gnupg.org/documentation/manuals/gpgme/Generating-Keys.html). 

I would like to use these features but currently hombrew does not allow installing `gnupg21` alongside `gpgme`.


MikeMcQuaid:
Do they both `brew link` side-by-side without problem?

jeroenooms:
Yeah I think so, no trouble so far.

MikeMcQuaid:
@jeroenooms I'm not sure how you got this to work given `gpgme` depends on `gnupg2` and there's a `conflicts_with` that too...

jeroenooms:
I did something like this:
```
brew install gpgme
brew remove gpg-agent dirmngr gnupg2
brew install --force gnupg21
```

screen shot 2016-10-12 at 6 37 25 pm

jeroenooms:
Note that recent versions of Debian/Ubuntu have no fully switched to GnuPG 2.1 in combination with GPGME 1.7 as well, if that is any precedent.

MikeMcQuaid:
@jeroenooms That is good precedent, thanks. I think we'll pass on this PR as-is as it requires too many workarounds but the best next step would be to submit one to upgrade to 2.1 in homebrew/core. Thanks!

jeroenooms:
Do you recommend to move `gnupg21` to `homebrew-core`, or simply bump `gnupg2` to version 2.1?

MikeMcQuaid:
Let's try the latter.

PR #1433, opened 2016-10-26
RFC: rpm4 (new formula)

petere:
This is the original rpm software, used by Red Hat.  The rpm formula in homebrew-core provides the "rpm5" variant.

I have been working with upstream for a few months to get this rpm package to build cleanly on macOS. The master branch now builds cleanly, but there is no tagged release of that yet. Upstream works in bursts, so it might be a while. In the meantime, I would rather not let this work rot on my machine. What should I do?
- Let this PR linger until a suitable release is tagged?
- Submit to `homebrew-headonly` now and move later when a release is tagged?

Is `homebrew-versions` even the right place anymore, since this isn't really a version but more of a fork or variant? Should I plan on submitting to `-core` instead? Anyone for renaming rpm in core to rpm5?

Comments on the actual formula welcome, too, of course.

### DIFF
--- a/rpm4.rb
+++ b/rpm4.rb
@@ -1,0 +1,122 @@
+class Rpm4 < Formula
+  desc "RPM package manager"
+  homepage "http://rpm.org/"
+  #url "http://rpm.org/releases/testing/rpm-4.13.0-rc2.tar.bz2"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I think unless this is something you personally actively plan on using/needing it's worth just dying and/or creating your own dedicated tap for it.

MikeMcQuaid:

PR #1442, opened 2016-11-05
rebar3 3.3.2

rjcoelho:
Created with `brew bump-formula-pr`.


MikeMcQuaid:
Thanks again @rjcoelho!

PR #1443, opened 2016-11-07
Adding formula for Mediainfo 0.7.70

HaraldNordgren:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [x] Does your submission adhere to Versions "Acceptable Formulae" [guidelines](https://github.com/Homebrew/homebrew-versions/blob/master/README.md#acceptable-formulae)?

-----

This corresponds to the latest Mediainfo version on the `python:3.5` Docker image, when running `apt-get update && apt-get install -y mediainfo`.

I don't know how the sha256 is supposed to be generated. I got mine from running `brew create https://mediaarea.net/download/binary/mediainfo/0.7.70/MediaInfo_CLI_0.7.70_GNU_FromSource.tar.bz2 --set-name media-info0770`.


HaraldNordgren:
I am not totally happy with how Homebrew taps seem to work for versions.

I expected to be able to effortlessly run `brew switch media-info 0.7.70`, but instead I have to do,

```
brew unlink media-info
brew link media-info0770
```

and vice versa to go back.

HaraldNordgren:
Using the alias, files are installed to `/usr/local/Cellar/media-info/0.7.70` enabling us to use `brew switch` to go between version, instead of having to link and unlink manually.

media-info0770/media-info.rb
```diff
@@ -0,0 +1,44 @@
+class MediaInfo < Formula
```
MikeMcQuaid: This file should be named `media-info07` and be in the root with class name `MediaInfo07`.

Aliases/media-info0770
```diff
@@ -0,0 +1 @@
+../media-info0770/media-info.rb
```
MikeMcQuaid: This alias should be removed.

MikeMcQuaid:
> Does your submission adhere to Versions "Acceptable Formulae" guidelines?

This does not meet those guidelines as it's the same major/minor version as the current stable. Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

HaraldNordgren:
Could you point me to the part of the docs that state this?

This says the opposite:

Versions formulae must not exceed +/-2 major/minor (not patch) versions from the current stable release.

MikeMcQuaid:
This differs only in patch version so is not acceptable. How would you suggest we reword that so it's clearer?

HaraldNordgren:
I guess, explain that the version *has* to differ in major/minor version.

The reason I submitted this is because the Debian version is incompatible with the current Brew version. Things like codec_profile returns 'Main@L3' instead of 'Main@L3.0' and different values for counts.

HaraldNordgren:
Regarding the naming of the class + the symlink. I am aware that this is outsid of the convention of this repository. But it has big advantages.

I wanted to achieve a situation where `brew install homebrew/versions/media-info0770` installs the files into the same Cellar as regular media-info, (so you can use `brew switch`). But you still wouldn't be able to accidentally run `brew install homebrew/versions/media-info`.

I think we would benefit from reorganizing the rest of the formulae like this.

MikeMcQuaid:
We're moving to a different versions approach in https://github.com/Homebrew/brew/issues/620 which will have side-by-side formulae rather than using `brew switch`. Eventually this tap will be migrated to homebrew/core.

PR #1448, opened 2016-11-09
thrift 0.9.2

rooftopsparrow:
Simple copy from https://github.com/Homebrew/homebrew-core/commits/56d8c1eba1e5ac30290dd0c486f4bba37f821e42 which was the last update to thrift 0.9.2

### DIFF
--- a/thrift092.rb
+++ b/thrift092.rb
@@ -1,0 +1,88 @@
+class Thrift < Formula
+  desc "Framework for scalable cross-language services development"
+  homepage "https://thrift.apache.org/"
+
+  stable do
+    url "https://www.apache.org/dyn/closer.cgi?path=thrift/0.9.2/thrift-0.9.2.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
We don't accept changes based only on patch versions, sorry!

rooftopsparrow:
I would like to petition the closing based on version number alone because there are _significant_ changes for different languages based on just the patch version ( because Apache sucks at versioning ). 

Here is the [CHANGELOG](https://github.com/apache/thrift/blob/fd832242bba9d4eaba71c7c08e8eed440b106f98/CHANGES) for the thrift repo

MikeMcQuaid:
Please consider [creating your own tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md#how-to-create-and-maintain-a-tap) instead. Thanks!

PR #1452, opened 2016-11-12
cmake30 and cmake31: cleanup

iMichka:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/homebrew-core/blob/master/CONTRIBUTING.md) document?
- [ ] Does your submission adhere to Versions "Acceptable Formulae" [guidelines](https://github.com/Homebrew/homebrew-versions/blob/master/README.md#acceptable-formulae)?

-----

Move the depends to the top for both formulae as asked by brew audit
Removed qt dependency, which is already gone for cmake31 and the
latest cmake in core.
See also homebrew-core commit e427965e4ba95007de0a47679bd09f13fef2178d

### DIFF
--- a/cmake30.rb
+++ b/cmake30.rb
@@ -33,11 +33,14 @@ class Cmake30 < Formula
 
   option "without-docs", "Don't build man pages"
   depends_on :python => :build if MacOS.version <= :snow_leopard && build.with?("docs")
-
-  depends_on "qt" => :optional
+  depends_on NoExpatFramework
 
   conflicts_with "cmake", :because => "both install a cmake binary"
   conflicts_with "cmake31", :because => "both install a cmake binary"
+
+  # The `with-qt` GUI option was removed due to circular dependencies if
+  # CMake is built with Qt support and Qt is built with MySQL support as MySQL uses CMake.
+  # For the GUI application please instead use brew install caskroom/cask/cmake.
 
   resource "sphinx" do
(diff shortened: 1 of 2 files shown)


iMichka:
About the point "Does your submission adhere to Versions "Acceptable Formulae" guidelines?":
current cmake is 3.7, so we could probably remove cmake 3.0 and 3.1. In that case we need to add cmake 3.5 and 3.6 here. On the other side these formulae (3.0 and 3.1) have been modified one month ago; so this point was ignored anyway.

I am not sure why these old cmake versions are still maintained here?

The purpose of this patch is to help for the qt5 transition here: https://github.com/Homebrew/homebrew-core/issues/1705

MikeMcQuaid:
> I am not sure why these old cmake versions are still maintained here?

Basically we keep them (they aren't particularly "maintained") until someone goes to add a new version. This process may change in future.

Unfortunately given that this formula is not building on 10.10/10.11 it may be worth just removing this version entirely rather than try to figure out how to get it building there again. Regardless, thanks @iMichka for the PR and help.

iMichka:
Right, makes sense. I also thought we should remove them in the first place. Closing in favour of #1454.

If somebody really needs these formulae they can make a PR to add them back. But cmake has a quite stable API, they tend to be rather conservative and backward compatibility is good, so there is probably no reason to keep older cmake versions around.

PR #1462, opened 2016-11-20
gnupg21: re-enable "make check"

ilovezfs:


ilovezfs:
Nope, "FAIL: ssh.scm" is still happening.

